PR #14, opened 2015-05-01
Passing $Encoding to Export-Csv and Disposing ExcelPackage Object

schneider-netservices:


dfinke:
Thank your for catching that and adding it

PR #31, opened 2015-07-31
My way to change Cell Colors

chrisk59:
What do you think about it ?


dfinke:
Thank you  for the PR, I'll review. It may take a bit of time.

chrisk59:
Added Tables Styles

PR #35, opened 2015-08-21
Allow reading a file already open in Excel

francoislg:
I've had a use case where I needed a file to be read while opened in Excel. 

It's a simple change, I doubt it will break anything.

The only thing this does is open the file handle with FileShare flag set to ReadWrite instead of Read, so it doesn't throw "The process cannot access the file 'X' because it is being used by another process" when trying to Import-Excel it.


dfinke:
Thanks for the great update

PR #42, opened 2015-10-01
Update ImportExcel.psm1

irwins:
Created a ValidateSet for Encoding and Extension. Removed Extension from the params hashtable. Removed Encoding at the Export-Csv as it is already included in params


dfinke:
Thanks for the update

PR #48, opened 2015-10-20
Added help for Export-Excel

DarkLite1:
Improved help by adding a richer explenation of the different parameters
and better examples. Also note that 'TitleBold' is no longer of type
Boolean but is now a switch as it can only true or false.

### DIFF
--- a/Export-Excel.ps1
+++ b/Export-Excel.ps1
@@ -1,34 +1,178 @@
-function Export-Excel {
-    <#
-        .Synopsis
-        .Example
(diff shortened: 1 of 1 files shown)


dfinke:
Wow, fantastic! Thank you very much.

DarkLite1:
You're welcome Doug. For evertything you did for the community, this was the least I could do for you. It took me a while to figure out GitHub, but it's great to be able to help. Have a nice day :)

PR #62, opened 2016-01-07
Adding AllWorkSheets option to Import-Excel to return the complete document in one PSObject

gaelcolas:
I Needed to be able to import a full Workbook instead of a single sheet, if possible without iterating through worksheets until it failed.
To avoid breaking backward compatibility I implemented the option as another parameter set, while setting the default parameter set as the original function syntax.

I think the form
```Powershell
if($AllWorkSheets) { ... } 
else { ... }
```
is clearer/less verbose than 
```Powershell
switch ($PsCmdlet.ParameterSetName) {
         'AllSheets' { ... }
         'SingleSheet' { ... }
}
```

Let me know if you're happy with this quick hack, or if you have any suggestion.

### DIFF
--- a/ImportExcel.psm1
+++ b/ImportExcel.psm1
@@ -13,207 +13,207 @@ Add-Type -Path "$($PSScriptRoot)\EPPlus.dll"
 . $PSScriptRoot\Pivot.ps1
 
 function Import-Excel {
-            [cmdletBinding(DefaultParameterSetName='SingleSheet')]
-                param(
-                                [Alias('FullName')]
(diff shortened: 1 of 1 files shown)


gaelcolas:
There's a weird behavior that I just realized... Closing pull request.
Looks like it reads only the properties set in the first sheet for the other sheets. Sounds like some variable needs to be re-initialized in the foreach loop...
Sorry for the inconvenience

dfinke:
No problem. I really like the idea. Let me know if I can help

PR #80, opened 2016-03-28
Updated Title related parameters

jdhitsolutions:
Changed the TitleBold parameter to a Switch for a better PowerShell experience. I also added some logic so that if the user specifies a title background color and do not set a TitlePattern to something other than None, they will get a warning message, but the rest of the Excel workbook will be created.


dfinke:
Thanks Jeff, I think this also fixes and error too.

PR #81, opened 2016-03-30
Added ValidateScript block to Path parameter for Import-Excel

brianaddicks:
Currently, when you give Import-Excel an invalid path you get:

Cannot convert argument "1", with value: "Open", for "FileStream" to type "System.IO.FileAccess": "Cannot convert value "Open" to type "System.IO.FileAccess". Error: "Unable to match the identifier name Open to a valid enumerator name. Specify one of the following enumerator names and try again:
Read, Write, ReadWrite""

Took me a while to troubleshoot that down to me being stupid and not giving it a valid path.  This will make the Cmdlet error out before it ever runs.

### DIFF
--- a/ImportExcel.psm1
+++ b/ImportExcel.psm1
@@ -29,6 +29,7 @@ function Import-Excel {
     param(
         [Alias("FullName")]
         [Parameter(ValueFromPipelineByPropertyName=$true, ValueFromPipeline=$true, Mandatory=$true)]
+        [ValidateScript({ Test-Path $_ -PathType Leaf })]
         $Path,
         [Alias("Sheet")]
         $WorkSheetname=1,


dfinke:
Thanks Brian! Sorry for the stall. This will save others time.

Seeing this highlights that it is a best practice.

brianaddicks:
Yeah, the only issue with this is that the error you get when you provide
an invalid path is a little cryptic.  Depending on my audience sometimes
I'll just do a test-path inside the function that will Throw an easier to
understand message:

if (!(Test-Path $YourPath)) { Throw "Invalid path specified for YourPath" }

This module is awesome btw.  So much faster than working with the Excel Com
Objects.

On Wed, Mar 30, 2016 at 9:49 AM, Doug Finke <notifications@github.com>
wrote:

> Thanks Brian! Sorry for the stall. This will save others time.
>
> Seeing this highlights that it is a best practice.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/dfinke/ImportExcel/pull/81#issuecomment-203441864>
>



-- 
Brian Addicks

PR #83, opened 2016-03-30
Fix for #82: multiple series aren't supported in a single chart.

### LINKED ISSUE #82: Charts with multiple series doesn't seem to work
I was trying to add two series to one chart using the following code:
```
Import-Module -Name ImportExcel -Verbose
$raw = @"
A,B,C,Date
2,1,1,2016-03-29
5,10,1,2016-03-29
"@
$data = ConvertFrom-Csv -InputObject $raw
Remove-Item -LiteralPath C:\temp\temp.xlsx
$data | Export-Excel -Path c:\temp\temp.xlsx -AutoSize -TableName Impressions -Show -ExcelChartDefinition @(
    New-ExcelChart -Title Impressions -ChartType Line -Header "Something" -XRange "Impressions[Date]" -YRange @("Impressions[B]","Impressions[A]")
) 
```

But it doesn't seem to work, it would appear that ```XRange``` referenced on https://github.com/dfinke/ImportExcel/blob/master/Export-Excel.ps1#L359 should actually be ```YRange``` (as that is the property it indexes into in the loop: https://github.com/dfinke/ImportExcel/blob/master/Export-Excel.ps1#L365).

redoz:
This fixes #82 

Now a list of ranges passed as YRange results in multiple series in a single chart.

### DIFF
--- a/Export-Excel.ps1
+++ b/Export-Excel.ps1
@@ -356,7 +356,7 @@ function Export-Excel {
             $chart.SetPosition($chartDef.Row, $chartDef.RowOffsetPixels,$chartDef.Column, $chartDef.ColumnOffsetPixels)
             $chart.SetSize($chartDef.Width, $chartDef.Height)
 
-            $chartDefCount = @($chartDef.XRange).Count
+            $chartDefCount = @($chartDef.YRange).Count
             if($chartDefCount -eq 1) {
                 $Series=$chart.Series.Add($chartDef.YRange, $chartDef.XRange)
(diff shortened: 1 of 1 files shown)


dfinke:
This is great! Amazing how a single property name change makes it pop.

As you also discovered other issues, this uncovers changes needed in Invoke-Sum, the measures parameter needs to be an array.

Thanks, let me know if you need this sooner. 

I'm going to test, make some other updates and then publish.

Thanks
Doug

Issue #86, opened 2016-03-30
Cannot use StartRow and StartColumn in combination with TableName


redoz:
Repro:

```
Import-Module -Name ImportExcel
$raw = @"
A,B,C,Date
2,1,1,2016-03-29
5,10,1,2016-03-29
"@

$data = ConvertFrom-Csv -InputObject $raw

$data | Export-Excel -Path c:\temp\temp.xlsx -AutoSize -TableName Impressions -Show -StartRow 1 -StartColumn 4
```

Table ends up _only_ on column A.

Setting StartColumn to 5 results in the following error:

```
Exception calling "Add" with "2" argument(s): "Object reference not set to an instance of an object."
At C:\dev\ImportExcel\Export-Excel.ps1:253 char:13
+             $tbl = $ws.Tables.Add($targetRange, $TableName)
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : NullReferenceException

The property 'TableStyle' cannot be found on this object. Verify that the property exists and can be set.
At C:\dev\ImportExcel\Export-Excel.ps1:255 char:13
+             $tbl.TableStyle=$TableStyle
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : PropertyNotFound
```

PR #87, opened 2016-03-30
Fixed range calculation for creating tables with StartRow and StartColumn

redoz:
Fix for #86

### DIFF
--- a/Export-Excel.ps1
+++ b/Export-Excel.ps1
@@ -245,8 +245,8 @@ function Export-Excel {
 
             $csr=$StartRow
             $csc=$StartColumn
-            $cer=$ws.Dimension.End.Row #-$StartRow+1
-            $cec=$script:Header.Count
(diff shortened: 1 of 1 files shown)


redoz:
I don't think this fix is complete, running:
```
Import-Module -Name C:\dev\ImportExcel\ImportExcel.psm1 -Verbose -Force
$raw = @"
A,B,C,Date
2,1,1,2016-03-29
5,10,1,2016-03-29
"@

$data = ConvertFrom-Csv -InputObject $raw
Remove-Item C:\temp\temp.xlsx
$data | Export-Excel -Path c:\temp\temp.xlsx -TableName One -StartRow 5 -StartColumn 8
$data | Export-Excel -Path c:\temp\temp.xlsx -TableName Two -Show 
```
Results in a corrupt xlsx.

redoz:
Closed, I decided to use http://github.com/redoz/ExtraLarge instead

PR #90, opened 2016-04-13
'Import-Module ImportExcel' no longer fails for PowerShell 4. Added support for Pester unit tests.

W1M0R:
When I did a pull of ImportExcel, my PowerShell 4 module no longer worked, because version 4 doesn't understand classes. A PowerShell 4 solution might require the use of Add-Type with a CSharp specification for the PSPlot class.


dfinke:
Thanks for the PR I will merge later. I added the v5 classes to demo two things, classes and how you can create a fluent API with PS. 

I think your check makes the most sense.

W1M0R:
I wanted to create a New Pull Request, but looks like Github preferred to reuse the existing pull request. Sorry about that. In this new commit, I added support for Pester unit tests, and the initial ExportSimple test that verifies the default behavior of Export-Excel.

dfinke:
Thanks @W1M0R! This looks great. I won't be able to review this till the weekend. I'll let you know if I have comments.

This module sorely needs some Pester love.

W1M0R:
In the last two commits, I have updated the unit tests, made corrections to the PS version test to be compatible with PowerShell 5, and added ConvertTo-TypedObject as a way to create a PsCustomObject with strict typing that can be passed to Export-Excel. I am no going to try and correct the leading zero issue (allowing the user to opt-in with a switch or something).

dfinke:
Thanks @W1M0R for this. These are too many changes at once for me to review at the moment. They're all great. I need to run these locally to get a sense of what they do and their impact. Perhaps you could look into how to create separate PRs for each? Maybe another branch on your side.


I like the separate files for each. I think I'd  prefer the `Test-ImportExcel` in as separate ps1 and not exported in the psm1.

What do you think?

Thanks for the great work. Hopefully I'll free up some time, these are great additions!

W1M0R:
The latest commit now makes it possible for users to export numeric strings with leading zeroes as a string and not as a number. I also added tests to confirm this behaviour.

Thanks for the feedback. I will do as you requested.

PR #91, opened 2016-04-15
Adds switch for PivotDataToColumn

pauldalewilliams:
By default, calculated values for PivotTable data appears on rows.  This switch changes that behavior so the calculated values appear on columns instead.

### DIFF
--- a/Export-Excel.ps1
+++ b/Export-Excel.ps1
@@ -27,6 +27,7 @@ function Export-Excel {
         [string[]]$PivotRows,
         [string[]]$PivotColumns,
         $PivotData,
+        [Switch]$PivotDataToColumn,
         [string]$Password,
         [OfficeOpenXml.Drawing.Chart.eChartType]$ChartType="Pie",
         [Switch]$IncludePivotTable,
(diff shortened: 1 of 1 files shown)


dfinke:
Very cool! Just to confirm. This is the expected result

```powershell
$f="c:\temp\test.xlsx"

rm $f -ErrorAction Ignore

ps |
    ? Company |
    select Company, Handles, PM |
    Export-Excel $f -Show `
        -IncludePivotTable `
        -PivotRows Company `
        -PivotData @{Handles='sum'; PM='sum'} `
        -PivotDataToColumn
```

image

pauldalewilliams:
Yes, that's what I was going for.  If you run it without the switch, do you get paged memory and virt memory as sub-rows of each company?  That was the behavior I was seeing before implementing the switch.

dfinke:
Yes. This is great! I've wanted to add this feature and did not have time to research it. I should have time to merge it today.

Thanks!
Doug

pauldalewilliams:
Glad I could help!  Yeah, I was just fixing it manually but decided there had to be a way to fix it programmatically.

PR #94, opened 2016-04-18
Added compatibility check against PowerShell 4.

W1M0R:
Again, sorry for the dirty commit history.


dfinke:
Thanks for all this. It's going to take me time to go through it. I feel much more comfortable that these functions are separated out. Unfortunately I don't have unit tests, so manually testing could take a bunch of time. 

It looks like there are only a couple of changes to the core though. The check for v4 so the classes are not loaded and the the `-StrictTyping` switch.

I don't see tests for the code for that switch. That's the challenge with this, needing to create actual xlsx files and then opening to inspect the data to see if the results are correct.

W1M0R:
The "restored branch to a state before I made changes commit" undoes all that other stuff. This pull request only has the version checking. The other pull request only has the test functionality. I pulled my hair out with git. If you Google for "cleaning up dirty pull request" you can get rid of my confusing commits and rollback strategy. 

You can rest in peace, the StrictTyping functionality is not included in this PR. I'm playing around with that idea, writing unit tests along the way.

W1M0R:
I'll close this request and try cleaning things up some more.

PR #96, opened 2016-04-19
Added support for PowerShell version checks.

W1M0R:
Previously, when the ImportExcel module was imported into a PowerShell 4
session, the import would fail. This functionality solves that problem.
To confirm this, cd into ImportExcel and run 'Import-Module .\ -Force'.
If you are running in a PowerShell 4 session, then a warning will be
displayed.

### DIFF
--- a/ImportExcel.psm1
+++ b/ImportExcel.psm1
@@ -16,7 +16,13 @@ Add-Type -Path "$($PSScriptRoot)\EPPlus.dll"
 . $PSScriptRoot\Get-HtmlTable.ps1
 . $PSScriptRoot\Import-Html.ps1
 . $PSScriptRoot\Get-Range.ps1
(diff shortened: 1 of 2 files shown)


dfinke:
Just and update. I'm not ignoring this. This is great work and terrific addition. I'm working out so I can dedicate the right amount of time to review this weekend.

Thank you

Doug

W1M0R:
No problem! Thanks for the great work you are doing.

dfinke:
A few more tweaks were needed. The New-Plot function had to be part of the version check as well. Also, I preferred a simpler way to check the version.

https://github.com/dfinke/ImportExcel/commit/045234dd81c8e5f482f3d0a5d3feafc516267d69

Thanks for the PR, great idea.

W1M0R:
Thanks!

PR #99, opened 2016-05-01
Made test independent of module path.

W1M0R:
By specifying the path, instead of the module name, the pester test can run without having the module installed into the global/user module path.

### DIFF
--- a/Export-Excel.Tests.ps1
+++ b/Export-Excel.Tests.ps1
@@ -1,6 +1,6 @@
 # Contributed by https://github.com/W1M0R
 
-Import-Module ImportExcel -Force 
+Import-Module $PSScriptRoot -Force -Scope Global
 
 function New-TestWorkbook {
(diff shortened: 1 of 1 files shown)


dfinke:
I need to think about this. It'd be great as a switch. I'd prefer to run it against my module path because it can uncover issues as it would be on a target users machine.

W1M0R:
Another option is to remove the Import-Module statement entirely. The user is then responsible for importing the module before running Invoke-Pester, which is probably what they will do in any case. Another advantage of removing the Import-Module statement from this, and other .Tests.ps1 files, is that tests will execute faster. In my case, I wanted to run Invoke-Pester without importing the module, without putting the module in my global or user module path, and without updating my $Env:PSModulePath variable. The $PSScriptRoot solution made that possible.

DarkLite1:
The `Pester` tests have been updated (#202) in the meantime taking care off this issue. So IMHO this PR can be closed as solved.

PR #104, opened 2016-05-20
Update README.md Installation Instructions

JustinGrote:
Update the installation instructions to be more direct and precise, making it easier for someone to get started. Also add the -scope CurrentUser method as an option which doesn't require admin privileges.


dfinke:
Thanks!

Issue #162, opened 2017-02-07
Import-Excel taking ages due to empty fields


DarkLite1:
UPDATE: Only read my last comment. it's faster then this stuff

We're having a specific xlsx file with a sheet called 'Permissions' that takes ages to read. We're still using an older version (v 1.97) of the Import-Excel module. So I did the same test with the newest version of your module to see if it fixes the problem, but the output is different..

**The test:**

`$F = 'C:\Temp\BEL MTX-STAFF-SCM.xlsx'`
`Import-Excel $F -Sheet 'Settings' -NoHeader`

> OUTPUT (Current version):

> 
>     P1 : Status
>     P2 : SiteName
>     P3 : SiteCode
>     P4 : ComputerName
>     P5 : Path
>     P6 : GroupName
>     P7 : Action
> 
>     P1 : Enabled
>     P2 : Supply Chain Management
>     P3 : SCM
>     P4 : DEUSFHEID0005
>     P5 : E:\DEPARTMENTS\SCM
>     P6 : BEL ROL-STAFF-SCM
>     P7 : Fix
> 
>     P1 : 
>     P2 : 
>     P3 : 
>     P4 : 
>     P5 : 
>     P6 : 
>     P7 : 
> 
>     P1 : 
>     P2 : 
>     P3 : 
>     P4 : 
>     P5 : 
>     P6 : 
>     P7 : 
> 
> OUTPUT (Old version):
> 
>     P1 : Status
>     P2 : SiteName
>     P3 : SiteCode
>     P4 : ComputerName
>     P5 : Path
>     P6 : GroupName
>     P7 : Action
> 
>     P1 : Enabled
>     P2 : Supply Chain Management
>     P3 : SCM
>     P4 : DEUSFHEID0005
>     P5 : E:\DEPARTMENTS\SCM
>     P6 : BEL ROL-STAFF-SCM
>     P7 : Fix

`Import-Excel $F -Sheet 'Permissions' -NoHeader`

> OUTPUT (Current version):
>     P1  : 
>       
>       
>           Matrix SCM 
>           R: Read
>           W: Write
>     P2  : Management
>     P3  : BPM logistics
>     P4  : Master data officer
>     P5  : RMC BE OID manager
>     P6  : RMC BE Team Coordinator
>     P7  : RMC BE Order intake
>     P8  : RMC BE Dispatch
>     P9  : RMC NL OID manager
>     P10 : RMC NL Team Coordinator
[...]

PR #163, opened 2017-02-08
Fix for #162 and other improvements

DarkLite1:
Fix for #162 
- Added help text for 'Import-Excel'
- Added 'DataOnly' parameter to 'Import-Excel'
- Improved 'Get-ExcelSheetInfo'

### DIFF
--- a/Export-Excel.ps1
+++ b/Export-Excel.ps1
@@ -1,17 +1,18 @@
 function Export-Excel {
     <#
-        .Synopsis
-        .Example
-        gsv | Export-Excel .\test.xlsx
-        .Example
-        ps | Export-Excel .\test.xlsx -show\
-        .Example
-        ps | Export-Excel .\test.xlsx -WorkSheetname Processes -IncludePivotTable -Show -PivotRows Company -PivotData PM
-        .Example
-        ps | Export-Excel .\test.xlsx -WorkSheetname Processes -ChartType PieExploded3D -IncludePivotChart -IncludePivotTable -Show -PivotRows Company -PivotData PM
-        .Example
-        Remove-Item "c:\temp\test.xlsx" -ErrorAction Ignore
-        Get-Service | Export-Excel "c:\temp\test.xlsx"  -Show -IncludePivotTable -PivotRows status -PivotData @{status='count'}
+        .SYNOPSIS
+            Export data to an Excel work sheet.
+        .EXAMPLE
+            gsv | Export-Excel .\test.xlsx
+        .EXAMPLE
+            ps | Export-Excel .\test.xlsx -show\
+        .EXAMPLE
+            ps | Export-Excel .\test.xlsx -WorkSheetname Processes -IncludePivotTable -Show -PivotRows Company -PivotData PM
+        .EXAMPLE
+            ps | Export-Excel .\test.xlsx -WorkSheetname Processes -ChartType PieExploded3D -IncludePivotChart -IncludePivotTable -Show -PivotRows Company -PivotData PM
+        .EXAMPLE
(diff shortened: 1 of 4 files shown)


dfinke:
Awesome PR, even help is in there! I'm reviewing, so far so good. Thanks!

DarkLite1:
Thanks Dough :) It would be great to have this in the master branch, That way we can update the module again from your code without worrying of breaking our current stuff. As we already have these changes in production. 

If you see possibilities to optimize the code. please feel free to do so. I did the best I could.

DarkLite1:
Seems logical to split it up. But then I would maybe suggest to refactor it to one function called'Get-ExcelInfo' with parameters 'Sheet' and 'Workbook'. To avoid duplicating code and keep things simple.

On the topic of adding headers, does it matter? As we only import data and not write it to a sheet.

dfinke:
For the splitting of the functions, do you have two or one function with different params. I prefer the separate functions, that way it is more easily discover it with `Get-Command`. Plus duplication can be eliminated with and "internal" function for shared operations.

Good question on headers. I think it ought to be consistent. I use `Import-Excel` to read sheets, operate/calculate details. Having the headers as data could throw off the counts on things.

What do you think?

DarkLite1:
Yes, if that is the desired behavior then I fully agree with you. Consistency is key! 

With regards to Get-Command you are right. It's easier to find that way.

dfinke:
So, would you be up for making those changes? I'm guessing this is the code you run in your shop? So would these changes break what you're doing?

DarkLite1:
Yup. already done and submitted. Nice catch though with the duplicate header entry :)

All fixed as requested:
- Seperate function for 'Get-ExcelWorkbookInfo'
- Added proper closing of the stream for 'Get-ExcelWorkbookInfo' and 'Get-ExcelSheetInfo'
- Fixed duplicate headers issue with 'DataOnly'

It's always great to have a second pair of eyes. I really appreciate your feedback Doug, keep it comming if you see more issues. Your other request to create a seperate function to open and close the stream can be done when this is accepted in your branck. So we can start with a clean sheet instead of mixing stuff up in one PR.

dfinke:
Awesome! That fix looks good. `Get-ExcelWorkbookInfo` works. I was pleasantly surprise when I look at the implementation that you implemented "piping". Excellent. I'll test a bit more and then merge.

Thanks again.

DarkLite1:
I wanted to do piping to with 'DataOnly' but couldn't figure it out properly.

dfinke:
I guess there are a couple scenarios there. Piping multiple xlsx files and looking at more than a single worksheet in a file and maybe across files

dfinke:
Thank you @DarkLite1, great additions. I've tested and merged your PR. When you get a chance, could you take a look at the Read me and let me know if I need  to update it.

Are you on twitter? If so, let me know your id.

Thanks again

DarkLite1:
Awesome! Appreciate the merge, we can now follow your master branch again :) But I first need to update a small piece of my coffee to use the new Get-ExcelWorkbookInfo function

No, I'm not such a Twitter fan. Some US diplomats are better at that than me ;) But you have my email and Skype if you want to get in touch.

dfinke:
Hey @DarkLite1, just curious if you were able to update to the master branch and if you had any issues?

No rush.

Thanks

DarkLite1:
Yes, I've updated the code to the master branch since 3 days a go I believe. All is still running smooth, no issues so far. Thanks for asking.

dfinke:
Great! Thanks for adding the features and testing it live

DarkLite1:
No problem! We use your code extensively, so if something would be wrong it would pop-up quite fast AND I would let you know asap ;)

Issue #165, opened 2017-02-13
Export-Excel: Spaces not allowed in TableName


DarkLite1:
Consider the following code:

```
$ExcelFile = $env:TEMP + '\test.xlsx'
Remove-Item $ExcelFile -EA Ignore

$ExcelParams = @{
    Path          = $ExcelFile
    AutoSize      = $true
    FreezeTopRow  = $true
    TableName     = "Tablename with spaces"
    WorkSheetname = "Test sheet"
}
Get-Process | Export-Excel @ExcelParams
```

This outputs the following errors:
```
Exception calling "Add" with "2" argument(s): "Tablename has spaces"
At C:\Program Files\WindowsPowerShell\Modules\ImportExcel\Export-Excel.ps1:231 char:13
+             $tbl = $ws.Tables.Add($targetRange, $TableName)
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : ArgumentException
 
The property 'TableStyle' cannot be found on this object. Verify that the property exists and can be set.
At C:\Program Files\WindowsPowerShell\Modules\ImportExcel\Export-Excel.ps1:233 char:13
+             $tbl.TableStyle=$TableStyle
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : PropertyNotFound
```

When using the previous EPPLus.dll this wasn't a problem. But recently I updated the EPPlus.dll with the new one from the master branch and this problem pops up. The spaces in TableName is not a problem in the PowerShell code but more in the EPPlus.dll library I guess. 

Thank you for checking this.

DarkLite1:
This is an exception thrown by the EPPLus library. see line [120 here](https://github.com/DHMechatronicAG/EPPlus/blob/master/EPPlus/Table/ExcelTableCollection.cs).:

        private void ValidateTableName(string Name)
        {
            if (string.IsNullOrEmpty(Name))
            {
                throw new ArgumentException("Tablename is null or empty");
            }

            char firstLetterOfName = Name[0];
            if (Char.IsLetter(firstLetterOfName) == false && firstLetterOfName != '_' && firstLetterOfName != '\\')
            {
                throw new ArgumentException("Tablename start with invalid character");
[...]

PR #167, opened 2017-02-20
Fix for Export-Excel #165

DarkLite1:
Fixed handling of the TableName with parameter validation, we now throw
an error in case the TableName:
- starts with something else then a letter
- is NULL or empty
- Contains spaces
Just like in the orriginal EPPlus.dll library.

Added Try/Catch clause for the End clause. So errors like 'TableName is
not unique' are terminating errors and the function isn't executed any
further.

Also cleaned up for readability some cosmetic stuff:
- 'if(' to 'if ('
- double quotes to single quotes where needed

### DIFF
--- a/Export-Excel.ps1
+++ b/Export-Excel.ps1
@@ -1,15 +1,20 @@
-function Export-Excel {
+Function Export-Excel {
     <#
         .SYNOPSIS
             Export data to an Excel work sheet.
+
         .EXAMPLE
-            gsv | Export-Excel .\test.xlsx
+            Get-Service | Export-Excel .\test.xlsx
+
         .EXAMPLE
-            ps | Export-Excel .\test.xlsx -show\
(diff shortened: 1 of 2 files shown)


DarkLite1:
So Dough, do you think this can be accepted in your branch?

dfinke:
I just need to test it and have a closer look. From what I see, it should be good to go. I should have time by the weekend.

dfinke:
Thanks. This is great. The verbose messages worked out well too.

Doug

DarkLite1:
Agree, it makes things a bit more clear with regards to what is happening behind the scenes. I usually do this all the time in my code, adding verbose messages. When something fails or something isn't quite right, I run it with the `-Verbose` switch. Often this has been a great help to me to pinpoint the problem really fast. Thanks for merging buddy :+1:

PR #187, opened 2017-05-15
Major update to 'Export-Excel'

DarkLite1:
Added parameter ‘NoNumberConversion'
- Allows us to pass on data to Excel without trying to parse it as a number
Some code clean-up:
- Removed repeating code by using functions
- Some syntax clean-up for better readability
- More clear verbose messages
- Improved error handling
Improved help
- Added parameter explanations
- Added more examples

### DIFF
--- a/Export-Excel.ps1
+++ b/Export-Excel.ps1
@@ -417,8 +417,8 @@ Function Export-Excel {
                 else {
                     foreach ($Name in $script:Header) {
                         $ws.Cells[$Row, $ColumnIndex].Value = $Name
-                        $ColumnIndex += 1
                         Write-Verbose "Cell '$Row`:$ColumnIndex' add header '$Name'"
+                        $ColumnIndex += 1
                     }
                 }
             }
--- a/Export-Excel.ps1
+++ b/Export-Excel.ps1
@@ -1,13 +1,151 @@
 Function Export-Excel {
     <#
         .SYNOPSIS
-            Export data to an Excel work sheet.
+            Export data to an Excel worksheet.
+
+        .DESCRIPTION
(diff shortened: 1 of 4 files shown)


dfinke:
I was just thinking this weekend, wonder where @DarkLite1 has been, now I know!

Thanks for the updates!
 Of course because there are no Pester tests. This will be a big challenge to verify it all works.

DarkLite1:
:) I've tested plenty, that's why I also added some examples to run through. Because in the beginning I missed quite a lot but now I think it's production ready as non of the existing functionality is broke.

Please let me know what you think. It would surprise me if you found something. But no one is perfect :P

dfinke:
Well do. I'll let you know. Thanks, and thanks for all the work.

DarkLite1:
I have to add, it's now a couple of days in production here in the shop and all is fine. We're not using all the bells and whistles but still, I'm quite confident nothing is broken. Anyhow, thinking about implementing a 'Transpose' switch. We currently have a project where we need to export objects to Excel in rows instead of columns. Difficult to explain but transpose in Excel switches it perfectly.

But I'm waiting for your acceptance in the branch first before adding this feature.

dfinke:
That's good to hear. I was hoping to test it out this weekend. Not sure I'll have the time. I've wanted a transpose for a while, that'd be great.

DarkLite1:
That would be great if you could do those test. I've never used Pester or anything, but in the meantime it might be a good idea to put some of your tests in the .EXAMPLE part of the help. So we can run them on new functionality to see if all still works as designed.

Transpose is definitely coming! I'm gonna look into this..

dfinke:
Wow, @DarkLite1 it's been a month. Going to carve out time today to review. I'm using the latest vs code an powershell extension, just released. Should make it easier to find any issues.

Thanks again for the hard work.

dfinke:
Could you copy paste the top if this PR you did and add it to the README? Also, could you say a bit more in the example/doc about `-NoNumberConversion *` and how to use? 

Also, wondering if that should be the default for  `-NoNumberConversion` so it just works?

Thanks for all the work, fantastic!

PR #195, opened 2017-06-29
Update and rename ConvertTo-ExcelXlsx.ps1 to ConvertToExcelXlsx.ps1

NordbergKMD:
File renamed to ConvertToExcelXlsx.ps1
Process block added


dfinke:
Great, thanks!

PR #200, opened 2017-07-13
Import-Excel: Split multiline cell lines by newline character

bmsimons:
An Excel cell can contain multiple lines in one cell. Therefor I think it would be functional to return values separated by a newline character to an array, just as I would expect. Right now 'Import-Excel' would only return multiline cell values as a string, with this pull request multiline cell values are now returned as an array.

### DIFF
--- a/ImportExcel.psm1
+++ b/ImportExcel.psm1
@@ -114,7 +114,7 @@ function Import-Excel {
                     $newRow = [Ordered]@{}
                     
                     foreach ($C in $ColumnReference) {
-                        $newRow."P$($C.NewColumn)" = $worksheet.Cells[($_.Name),($C.Column)].Value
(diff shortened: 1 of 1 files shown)


DarkLite1:
Thanks @bmsimons for the great PR! This seems indeed to be a good idea. It would indeed make sense to create an `array` of values in case there are multiple lines in one cell. Would you be willing to re-apply your correction to the current branch/file? We had some big changes in the meantime and we would make merging a lot easier. Also, please add a small `Pester` test in the file `ImportExcel.Tests.ps1` that will confirm the new functionality. It will help us in deploying only code that improves and ie doesn't have regressions. Otherwise I'm afraid we'll have to close this PR as merging is currently not possible.

PR #202, opened 2017-08-01
Rewrote 'Import-Excel' to fix some serious bugs and added function 'Update-FirstObjectProperties' 

DarkLite1:
Rewrote 'Import-Excel':
- Added parameter sets for proper parameter validation and to make sure '-NoHeader' and '-HeaderRow' aren't used together
- Added try/catch clause, CmdLetBinding and verbose messages
- Renamed 'HeaderRow' to 'TopRow' to avoid confusion with other parameters
- Renamed '-Header' to '-HeaderName'
- Added test for duplicate property names
- Added test for empty worksheet
- Added test for no data after TopRow
- Fixed incorrect import when there's no value in the first column
- Fixed values being imported under the wrong property name in case one 
- Fixed incorrect import in case column A is empty and B and C not ( '$Worksheet.Dimension.Columns' is unreliable because it will say 2 columns are in use while it should say 3).
 (Ex. Add data in cell B2 and C2, use the '-NoHeader' switch, notice P1 and P2 are incorrectly blanc.)

- Added the function `Update-FirstObjectProperties.ps1`
- Improved some code for readability
- Added some examples in `Export-Excel`


DarkLite1:
Functionality hasn't changed, only added some `try/catch` clauses and a `Param()` clause to the installation files. Also changed `echo` to an `array` and ordered the list of `.ps1` files to download. Some where missing.. So all in all, nothing should be broken.

DarkLite1:
My improvements to `Import-Excel` are apparently immediately added to the first pull request, so hence this message. There were some serious bugs in `Import-Excel` that forced me to redesign the function completely. Feel free to test and accept/decline it.

PR #217, opened 2017-09-20
Import-Excel parameter

DarkLite1:
- Fixed position of `-WorksheetName` parameter
- Added extra Pester tests for parameters

### DIFF
--- a/ImportExcel.Tests.ps1
+++ b/ImportExcel.Tests.ps1
@@ -15,6 +15,9 @@ $Path = 'Test.xlsx'
 Context 'input' {
     in $TestDrive {
         Describe 'parameters' {
+            BeforeEach {
+                Remove-Item ./* -Force
+            }
             Context 'mandatory in sets' {
                 it 'Path' {
                     (Get-Command Import-Excel).Parameters['Path'].Attributes.Mandatory | Should be $true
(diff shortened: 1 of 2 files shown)


DarkLite1:
@dfinke if you accept this one it will be possible to omit the parameter names of `-Path` and `-WorksheetName`. So I think it's quite nice to have this because it makes the following just work:

`Import-Excel $Path $Sheet`

dfinke:
Tests pass, good stuff. 
Couple of things, I tried `Import-Excel` on a work box, it worked without specifying the param names. It was an older version but did not check it.

Also, I didn't have the `Assert.psm1` on my box, an the `# requires` caught it, cool. Wonder if that was what caused the issue on appveyor, doubt it, just a data point.

DarkLite1:
On my version of PowerShell, `5.1` it didn't work without setting the parameter `Position=1`. That's why I pushed this PR to fix that.

Yeah, I thought that the `#requires` would be handy to have on top. The `Assert` module is really cool! It has an experimental feature to compare objects (arrays). That's what we use for our tests. Makes life a bit easier and makes it more clear what is the `$Result` and the `$ExpectedResult`.

Really good stuff from the `Pester` guys. Use it from now on in all my production scripts too. Should've done that a long time ago :P

DarkLite1:
Maybe we should do a version bump to `4.0.1` so it's picked up by the `PS Galery`?

dfinke:
When there are more substantial updates, it'll be good to publish. Please, could you update the readme with a sentence on the parameter changes.

DarkLite1:
I understand. I don't think an update of the `README` is needed, because it was a bug as reported by a user. It was also clear from the examples that this was the expected behavior. It would just say that I made a mistake, and I don;t like admitting that ;)

PR #237, opened 2017-10-21
My changes to support append, and formatting

jhoneill:
I changed Export-Excel.ps1 as follows. 

#1 @ Line 197 Made new parameter sets . Default, and table already
existed and use path. Added DefaultPackage and TablePackage
A New parameter  "Package" allows an ExcelPackage object returned by
-passThru to be passed in
~Line 400 code to use package or path depending on path passed.

(also added Open-ExcelPackage  to get the object without exporting and
Close-ExcelPackage to close it nicely - these are in their own file)

#2.  @ Line 256 added new parameter excludeProperty to remove unwanted
properties without needing to go through select-object
~Line 459 added logic to exclude the properties specified in the new
parameter

#3 .  @ Line 262 Added new parameter Append
~Line 420 code to read the existing headers and move the insertion point
below the current data
(normal behaviour is to check if headers exist when adding data in the
process block, which makes this change wonderfully easy)
~Line 510 changed basis for identifying named ranges and changed scope
for rangeName so it can be used on other sheets

#4.  ~Line 550. Remove any existing Pivot table before trying to
[re]create it.

Added formatting.ps1 which applies conditional and normal formats -
requires an ExcelPackage to be open.

Added Open-ExcelPackage.ps1 (which contains a close function as well to
get the the object and save it )  open allows the sheet to be loaded
into a package object without needing to export .

Updated .psm1 to add the formating and open/close ps1 files.

### DIFF
--- a/Export-Excel.ps1
+++ b/Export-Excel.ps1
@@ -5,13 +5,83 @@
 
         .DESCRIPTION
             Export data to an Excel file and where possible try to convert numbers so Excel recognizes them as numbers instead of text. After all. Excel is a spreadsheet program used for number manipulation and calculations. In case the number conversion is not desired, use the parameter '-NoNumberConversion *'.
-
+        .PARAMETER Path
+            Path to a new or existing .XLSX file
+        .PARAMETER  ExcelPackage 
+            An object representing an Excel Package - usually this is returned by specifying -Passthru alllowing multiple commands to work on the same Workbook without saving and reloading each time. 
+        .PARAMETER WorkSheetName 
+            The name of a sheet within the workbook - "Sheet1" by default
+        .PARAMETER ClearSheet
+            If specified Export-Excel will remove any existing worksheet with the selected name. The Default behaviour is to overwrite cells in this sheet as needed (but leaving non-overwritten ones in place)
+        .PARAMETER Append
+            If specified data will be added to the end of an existing sheet, using the same column headings. 
+        .PARAMETER TargetData
+            Data to insert onto the worksheet - this is often provided from the pipeline. 
+        .PARAMETER ExcludeProperty
+            Speficies properties which may exist in the target data but should not be placed on the worksheet 
+        .PARAMETER Title
+            Text of a title to be placed in Cell A1
+        .PARAMETER TitleBold
+            Sets the title in boldface type
+        .PARAMETER TitleSize
+            Sets the point size for the title
+        .PARAMETER TitleBackgroundColor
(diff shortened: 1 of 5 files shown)


dfinke:
Awesome PR! Thanks, will review and merge!

Thanks again.

jhoneill:
I've added some more bits ... my GIT expertise is a bit lacking but I think it all becomes part of the same pending pull request.  

the new bits are 
Export-Excel.ps1
1. No code seems to act on the "NoClobber" parameter, I had previously added a "ClearSheet" parameter which removes the worksheet and any past data (before if there were fewer rows or columns in the new data, any old data would not be over-written)
2. I have made the whole of the process{} block in Export-Excel subject  to IF ($target data)  .... this way things which are processed in the end{} block can be done in a separate call to Export-Excel without the
need to load data.  Because of this there should be no need for the Open and Close -ExcelPackage functions going forward, but I've left them in place for now
3. Fixed a bug when setting the target range it was previously setting the end of the range to be the number of columns which failed if the data was inserted at a column other than one.
4. I've moved the top left corner of a pivot chart so it sits closer to the data.
5. I've changed the check for inserting a pivot table so if  -InsertPivotChart is specified it implies -insertPivotTable.

Formatting.
6. The previous check-in was not up-to date and contained some issues with conditional formatting which I had fixed in the past.
7. Fixed a peformance bug in set-format where a foreach intended to allow the same format to be applied to multiple ranges would format a large block of cells one-by-one if it was the only range passed. 
8. Improved online help for the formatting commands.

Export charts
9. My current project is importing a lot of data into Excel and needs to take Excel charts out as JPG files. I've thrown in the Export-Charts script as "something for the pot",  even though the connection with the
rest of work is loose :-) 

Best etc. 
J

dfinke:
Cool on the `Export-Charts`. I keep attacking that from the EPPlus side and didn't think of using COM.

jhoneill:
:-) I **think** the XLSx contains the XML to describe how to draw the chart but not the rendering of a drawn one, so I assumed I had to do that way. The pity is the script runs on a server with no excel and to get the graphs out the file needs to be copied to a client with Excel on it.

dfinke:
Checking out your updates, fantastic. Going to add some comments here.

Should `formatting.ps1` be renamed?

jhoneill:
-Now should probably be a parameter set on its own (or two: "Now" and "TableNow" 
formatting.ps1 can be renamed,  it's only called that because it has more than one command and the argument completer for color ... they could be split, merged into another file or kept in their own file under a new name.

PR #248, opened 2017-11-12
EPPlus supports Range.Indent this adds support for it

PetriAsi:
This adds support for indenting ranges with Set-Format
EPPlus library already supports this with indenting values between 0-15.

This was my fist github pull request, hope that I made it right.


dfinke:
Nice feature. Please could you add a `.EXAMPLE` in `Export-Excel` and I will merge

PetriAsi:
I added simple example.

PR #250, opened 2017-11-15
Minor fixes, plus set-row. set column and Send-SQLData

jhoneill:
Export-Excel : 
* Fixed a typo in the message at line 373.  
* Now catch an attempt to both clear the sheet and append to it. 
* Fixed some issues when appending to sheets where the header isn't in row 1 or the data doesn't start in column 1. 
* Added support for more settings when creating a pivot chart. 
* Corrected a typo PivotTableName was PivtoTableName in definition of  New-PivotTableDefinition
Add-ConditionalFormat and Set-Format added to the parameters so each has the choice of working more like the other. 
Added Set-Row and Set-Column - fill a formula down or across. 
Added Send-SQLDataToExcel. Insert a rowset and then call Export-Excel for ranges, charts, pivots etc


dfinke:
Thanks! Huge additions

PR #271, opened 2017-12-21
Update Get-HTMLTable

tehfonz:


dfinke:
These are cool, could you please add an example for this and the `Import-Html` in the **Examples** folder.

Plus, either update the `README` or leave a comment, just a line about these additions. 

Feel free to leave a twitter id or blog link too.

Thanks for the PR!

PR #298, opened 2018-03-13
Export-Excel: Added support for [DataTable]

ili101:
* Can process DataTable much faster than regular data, On my test data performance increased from 40 Minutes to 26 Seconds.
* Automatically take "Tablename" from the [DataTable] and column Int/String type.
* Notice that you need to use "-TargetData $MyDataTable" or add a comma if you pipe the variable ", $MyDataTable | Export-Excel" so PowerShell will not break up/enumerate the DataTable.

* Added Parameter -AutoSizeFirst [Int], It’s the same as -AutoSize but let you limit the number of checked lines to improve performance.

* Added Parameter -Table [Switch], To enable creating a Table without setting -TableName so name is default (Table1…) or with name from the [DataTable]

* Added DataTable Tests in Export-Excel.Tests.ps1

* Set 'thousand seperator and decimal number' test in Export-Excel.Tests.ps1 to -Pending, It's giving me an error, please check it, I think the test is wrong or is working only on specific regions or versions?

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,19 @@ Install-Module ImportExcel -scope CurrentUser
 ```PowerShell
 Install-Module ImportExcel
 ```
+
+# What's new 6.0.0
+
+[Get the zipped preview here](https://ci.appveyor.com/project/dfinke/importexcel/builds/23647194/artifacts)
+
+* Performance improvement to `Export-Excel` see [#506](https://github.com/dfinke/ImportExcel/issues/506) and [#555](https://github.com/dfinke/ImportExcel/issues/555). This has meant taking code in Add-CellValue back into process block of `Export-Excel`, as the overhead of calling the function was a lot greater than time executing the code inside it. [Blog post to follow](https://jamesone111.wordpress.com/). Some tests are showing a ***~10x*** speed increase. [#572](https://github.com/dfinke/ImportExcel/issues/572) was about a broken #region tag in this part of the code and that has been cleaned up in the process.
+
+* `Export-Excel` now has an -InputObject parameter (this was previously -TargetData , which is now an alias for InputObject).
+If the `inputobject` is an array, each item will be inserted, so you can run `export-excel -inputobject $x` rather than `$x | Export-Excel`, and if it is a `system.data.datatable` object it will be inserted directly rather than cell-by-cell. `Send-SQLDataToExcel` takes advantage of this new functionality. There are simple tests for these new items
+
+* `Export-Excel` previously assumed `-Now` if there were no other parameters, it will now assume `-Now` if there is no `-Path` or `-ExcelPackage`.
+The .PSD1 file now itemizes the items exported by the module [#557](https://github.com/dfinke/ImportExcel/issues/557)
+
 # What's new 5.4.5
 
 Thank you to [James O'Neill](https://github.com/jhoneill) for the great additions.
@@ -98,7 +111,7 @@ Thank you to [James O'Neill](https://github.com/jhoneill) for the great addition
 
 # What's new 5.3.4
 
-- HotFix for parameter PivotTableSyle should be PivotTableStyle https://github.com/dfinke/ImportExcel/issues/453
+- HotFix for parameter PivotTableStyle should be PivotTableStyle https://github.com/dfinke/ImportExcel/issues/453
 
 # What's new 5.3.3
 


dfinke:
Thanks for PR! A few things. Please do the `PSPlot` bug fix as a separate PR or I can do that. Also, it'd be great if you could add examples of using this in the `Examples` folder. Lastly, could you add the bullet points to the README.md. If you have other social media links (twitter, blog, etc) please add them there too.

Once the `PSPlot` is not here, I can take this for a test drive and then merge.

ili101:
I created a separated PS for PSPlot https://github.com/dfinke/ImportExcel/pull/316.
I will get some examples for this PR soon, I already added tests in there so it's just copying some of the tests code and adding some comments.

PowerShellToday:
Could I do anything to help to get this in production, I have been thinking about this for a while and really hope this will be available. I’m not that great in writing English, but maybe I could help out with some examples.
Another thought, not sure if it’s applicable is to have a switch-parameter that converts your data to a DataTable, or create a cmdlet that lets you convert ordinary arrays of data to DataTables and use that before sending data to Export-Excel. But I could start writing a Function ConvertTo-DataTable that take pretty much the same parameters as Export-Excel like TargetData, NoNumberConversion, ExcludeProperty that could be called either via a switch-parameter or by user as earlier step in the pipeline.

dfinke:
@PowerShellToday Yes. Once the 5.0 is tested and officially published. I'm working through how to create unit tests for this. Optimally that could speed the process to make this live.

ili101:
Sorry I didn't got to work on it yet. When I wrote this functionality I actually didn't know that the command Send-SqlDataToExcel existed. When I used it now I can see that it missing some of the functionality I added like formatting [DateTime] to Date in Excel and using the table name from the [DataTable] and my code missing some functionality that is in Send-SqlDataToExcel like the NoHeader.

So I think In order to not have duplicate code and inconsistent functionality it will be best to check that Export-Excel itself has all the [DataTable] functionality and use it in Send-SqlDataToExcel.

Other that that it work nicely I'm using it in production for some time without problems. I think I can get to finish this this week or next week.

Regarding ConvertTo-DataTable I can actually find at last 3 implementation of this on the web, do you want to use this in order to improve preformence? if so we need to test how fast they are (My data is already a DataTable so I don't need converting), alternatively we can probably also improve preformence (on non DataTable data) without converting to DataTable but use a list of Objects data type instead of the of the cell by cell method used now that probably just as fast as DataTable but do not requires conversion to DataTable. See the List example here https://github.com/JanKallman/EPPlus/issues/37#issuecomment-342342797, so maybe you can try to work on implementing that if your data is not a native DataTable.

PowerShellToday:
Hi @ili101,
Yes I was looking for performance of exporting large sets of data to Excel. And the way I saw it was to use the DataTable, but the link you provided opens up another way of doing this. And no I don’t have native DataTables as a start. So ConvertTo-DataTable was my solution to the problem, but I strongly believe you are correct that it may not be the "_correct_" way. Thanks for the suggestion, will start playing around with that and do some tests.

PrzemyslawKlys:
Since I essentially rewrote Export-Excel function in my pull request I wonder if this can be handled easier with new approach that I proposed. Which based on the data type is processed outside of main function and just passed back as array of arrays. https://github.com/dfinke/ImportExcel/pull/412

dfinke:
Several tests failed. Also, this PR contains code that will not be merged at this time. Please consider opening another PR with your updates, ensure all tests run and add more where necessary.

Thanks

ili101:
Hi @dfinke, I was not finished with the changes, just done for the day.
I finished updating the code and merging it with the updated version, also finished added the relevant tests.
All tests were passing locally so I Pushed to GitHub to see if it's passing in AppVeyor in the Pull request which there was a compatibility problem between the test fix and AppVeyor so one test was failing.
Today I refixed the test that should now pass in AppVeyor to, Tomorow I will start on the documentation.

dfinke:
Sounds good @ili101! Re-opening. Thank you for keeping an eye on the tests, and docs!

dfinke:
Interesting, Appveyor works and Azure DevOps has issues. Lets go with the success of Appveyor and I'll look into the Azure DevOps issues.

ili101:
Apparently I was using the "New" Pester syntax that is not supported on the old version that is pre-installed on Azure DevOps, they have info on it here https://github.com/pester/Pester/wiki/Migrating-from-Pester-3-to-Pester-4 and Azure DevOps update example mentioned here https://github.com/pester/Pester/wiki/Showing-Test-Results-in-CI-(TeamCity,-AppVeyor,-Azure-DevOps)

dfinke:
Cool, I the DoTests.ps1, I check to see if pester is installed. I'll update that to figure out versions as well.

ili101:
Hi
* I added CHANGELOG.md, you can move/edit its content to the appropriate place, I wasn't sure where to put it.
* Send-SqlDataToExcel.tests.ps1 don't including Database tests, will be good if someone will add some.
* The Export-Excel params copy in Send-SqlDataToExcel was stripped of theirs ParameterSets, and after adding them back I understand why. In order for it to work as intended you need to "multiply" the original Export-Excel ParameterSets with Send-SqlDataToExcel ParameterSets resulting in a lot of ParameterSets.
It works well but requires allot of ParameterSetNames, can you can think of a shorter way to do this?

PR #312, opened 2018-04-09
Add ReZip Parameter for MSOnline compatibility (fixes #240)

### LINKED ISSUE #240: Exported Excel doesn't import into PowerBI Online
When trying to import via the onedrive for business method, it returns "error 400" on the file. If I take the file, open it in excel, and save it, then it works fine. It also imports into PowerBI Desktop Fine without having to open it in excel first.

I wanted to use this in an Azure Function to batch export data from a proprietary system into a format I can use with PowerBI and automatically refresh hourly, and one excel with multiple worksheets is ideal so I can create relationships. I may be able to make this work with CSVs or an Azure SQL database I dump and rebuild periodicall but that is cumbersome.

Here is an example of "before" and "after" opening in excel and saving, maybe you can diff for what it's looking for. The excel file is significantly larger.

[testexport.xlsx](https://github.com/dfinke/ImportExcel/files/1419914/testexport.xlsx)
[testexport-saveinexcel.xlsx](https://github.com/dfinke/ImportExcel/files/1419915/testexport-saveinexcel.xlsx)

JustinGrote:


CrashM:
@JustinGrote WooHoo, I can confirm that this patch works.  I am able to send email with an Excel file created via Export-Excel with your patch through Microsoft Edge servers to Outlook 365.

dfinke:
Thanks @JustinGrote for the fix and testing. Thank you @CrashM for confirming.

Also, thanks for including the `Write-Verbose`

PR #316, opened 2018-04-16
Ps plot

ili101:
Fixed the bug "Unable to find type [PSPlot]


ili101:
I don't know if it the best way to do this but I think it worked
From the GitHub website I crated a branch from before my changes and updated it with the "compare" option.
Then copped the commit to it with "git cherry-pick e52731cd70c86b5e4f9addac2841404f82cb58e4"

PR #318, opened 2018-04-21
Replace empty string with NULL when generating SQL inserts. Fixes #288

### LINKED ISSUE #288: Blank value inserted as 0
I am using the ConvertFrom-ExcelToSQLInsert function to take multiple sheets in one Excel file to import in to SQL tables.  We found an issue yesterday where 16 values in a column on a sheet had blanks as the value, but when imported in to SQL it translated to a 0. (zero)

Is there a parameter I could use to prevent this behavior or is this a bug?

image

nzubair:
This PR attempt to provide a workaround for #288, where empty strings causes an INT column to be added with value of 0 instead of NULL. 

I've introduced a new parameter `-ConvertEmptyStringsToNull`. If specified, it will cause the resulting insert statement to have NULL instead of an empty string (`''`).

### DIFF
--- a/ConvertFromExcelToSQLInsert.ps1
+++ b/ConvertFromExcelToSQLInsert.ps1
@@ -1,4 +1,5 @@
 function ConvertFrom-ExcelToSQLInsert {
+    [CmdletBinding()]
     param(
         [Parameter(Mandatory = $true)]
         $TableName,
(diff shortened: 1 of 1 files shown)


dfinke:
Thanks @nzubair for the PR! I'm going to take this opportunity to add some Pester tests for this.

PR #389, opened 2018-07-12
Latest bunch of changes

jhoneill:
Details in readme 
Main changes - added more parameters to Join-Worksheet and  Send-SqlDataToExcel 
Added "quick chart" support to Export-Excel
Fixed and issue where Export-excel sometimes got confused in parameter sets (fix is to tell some params they can get their value via a pipeline property even though that never happens) 
Added to help text and made some more vs-code friendly. 
More and better tests


dfinke:
Curious, why did you remove the Get-ExcelColumnName?

I used that, infrequently, when I have a column number and convert it.

jhoneill:
Broke one of my own rules ( must have been working late :-( ) should not take out any commands which people might be using;  so I've backed that change out. 

The logic was that I was trying to flatten the code and use [OfficeOpenXml.ExcelAddress]::TranslateFromR1C1() 
 rather than carry a column letter and later mash strings to make an address. TBH I'm on the point of making a function to wrap translateFromR1C1 because it isn't very clear what it is doing.

dfinke:
All good, was wondering if you found a better way. I believe I ported that from C#, someone using VSTO.

On a version control note. I'm looking into the VSTS public repo offering. May port this there. I've heard you can have a single integrated build pipeline across OSes.

jhoneill:
Can no longer see the appveyor output :-(

Issue #397, opened 2018-07-20
Formating TextRotation only possible from 0 to 90 degrees anti-clockwise


CamFlyerCH:
I tried to rotate the contents of a cell clockwise. Against this modules parameter description an constraints the values accepted and used by EPPlus are not -90 to 90. They seam to be 0 to 90 for anti-clockwise, 91 to 180 for clockwise. (It contains 255 for vertical text, but it's "out of range" if I try to set that.

Modification proposal for the functions to format cells

            if ($TextRotation) {
                if ($TextRotation -lt 0) {
                    $Address.Style.TextRotation = 90 - $TextRotation
                } else {
                    $Address.Style.TextRotation = $TextRotation
                }
            }

dfinke:
Thanks! Would you want to try and do a pull request on that? Then it can be merged

PR #399, opened 2018-07-23
Fix für Issue #397 to convert TextRotation values to working values

CamFlyerCH:
The TextRotation parameter is described and tested for values from -90 to 90. But the values accepted (by the DLL ?) are 0 to 180. This works like this:

 0°   to   90°  =   Values 0 to 90
-1°   to  -90°  =   Values 91 to 180

I added a test to convert negative input values. While testing this I found that it was impossible to reset the text rotation to 0°. This is due the to simple test if a parameter is set by only do like
`if ($TextRotation){........}`

I changed this to 
`if ($PSBoundParameters.ContainsKey('TextRotation')) {.......}`
So if you specify now the option "-TextRotation 0" it resets the rotation to 0.

### DIFF
--- a/Set-Column.ps1
+++ b/Set-Column.ps1
@@ -123,7 +123,7 @@
     if      ($FontShift)               { $Worksheet.Column(     $Column).Style.Font.VerticalAlign          = $FontShift          }
     if      ($NumberFormat)            { $Worksheet.Column(     $Column).Style.Numberformat.Format         = $NumberFormat       }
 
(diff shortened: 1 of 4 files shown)


CamFlyerCH:
@dfinke Sorry, but I am still a novice in using GitHub. I think I will likely screw up you test code since I did not understand it properly.

dfinke:
Need to sort out the conflicts. The latest had a lot of refactoring.

Issue #400, opened 2018-07-23
Some formatting or unhideing is impossible


CamFlyerCH:
While testing the fix for Issue #397 I found that some of the parameter presence testing is bad or some function are impossible to reverse. This affects the following functions:

Set-Format: (and some parameters in Set-Row, Set-Column and Add-ConditionalFormatting too)
- -Hidden, is a switch parameter, but there is no unhide
- -WrapText, can only be enabled, but there is no disable
- -Bold, Italic, StrikeThrough and Underline, can only be undone all together with ResetFont, not seperatly
- -Width and -Height can't be set to 0
- -Value, can not be cleared

Export-Excel
-HideSheet , but no unhide

(This list is incomplete, but you get the point)

Converting switches to Boolean parameters would help, but this would be a breaking change.
Testing for used parameters would help to set 0 oder nothing. Example: 
`$PSBoundParameters.ContainsKey('MyParameter')`

jhoneill:
So a quick explanation of how we got to where we are : it was expected Export-Excel and set-format, set-column and Set-Row will be working on newly created worksheets, by default things will only be hidden, wrapped, bolded etc by one of these commands. So the slightly glib advice was (e.g.) if you don't want something hidden, don't hide it; don't set values you want cleared and so on.  I didn't foresee a need to turn wrapping off, clear values etc, and I don't think @dfinke  did either.  
I added Reset font for the case where I set two overlapping ranges and if one set bold and one set italic I wanted the second to be able to say "Italic and nothing else"   

Booleans and psBoundParameters are both problematic : consider 
```
function test {
param ([boolean]$b,
       [int]$i ,
       $j = 10 
       )
if ($b -eq $null)  {Write-Host 'b is null' } 
if ($b -eq $false) {Write-Host 'b is false' } 
write-host "Bound paramter i = '$($PSBoundParameters['i'])'"
write-Host "Variable i = $i" 
write-host "Bound paramter j = '$($PSBoundParameters['j'])'"
write-host "Variable J = '$j'"
}

test 
```

Most people would not expect b to return false ;  but it does (Powershell converts the Null value of B to a Boolean, to get round this you have to use a [nullable[Boolean]]) . 
A standard Boolean means that 
 `set-format … -bold $true ; set-format … -italic $true ` 
leaves bold unset, because the second command implies `-bold $false -underline $false` etc.

[...]

PR #405, opened 2018-07-27
Internal hyperlinks, reworked format parameters, unhide in export Excel

jhoneill:


dfinke:
@jhoneill not sure what is going on with the `readme.md` the last few times it has not picked up the changes in master

jhoneill:
In the Set-Row_Set-Column-SetFormat.tests.ps1 line 38 works on my machine but not in appveyor so I commented it out. 

I tried various things but couldn't insert this into the readme.md

# New to 25th July
- Added color parameter completer to a few places where it was missing (Add-ConditionalFormatting/PatternColor  New-ConditionalText PatternColor&BackgroundColor) 
- Changed charting.ps1 and examples\charts\*.ps1 to use New-ExcelChartDefinition instead of New-ExcelChart
- Quick charts in Export-excel were too wide (now 800 pixels instead of 1200), and now support ShowPercent, ShowCategory and NoLegend Parameters 
- Fixed bug in Add-ExcelChart where XAxisPosition and YAxisPostion would not be set correctly 
- Fixed bug in Set-Format where enums with a value of zero, or zero numbers would not be set; added functionality to set-format to support -bold:$false -italic:$false etc.  (see #400) 
- Changed HideSheet in Export-Excel to support wildcards, and added UnhideSheet. (also #400) 
- Added returnRange support to set-Column and Set-row
- Added tests for better coverage (now at >80% average - set-row/colum set-format less than 80%) , and tweaked some tests to use few rows and/or columns for speed

dfinke:
@jhoneill How did you determine code coverage? I was looking at tools.

jhoneill:
It's not very nice but
 `invoke-pester -CodeCoverage  "..\Export-Excel.ps1", "..\Join-Worksheet.ps1" ,  "..\compare-worksheet.ps1"  , "..\Merge-worksheet.ps1" , ..\Set-Column.ps1 , ..\Set-Row.ps1, ..\SetFormat.ps1, ..\AddConditionalFormatting.ps1   
`

dfinke:
Yeah, not a fan. Good in a pinch. Was looking at the code coverage approach in the PowerShell repo

dfinke:
Haven't seen that trick in a bunch of years :) `[System.Windows.Forms.SendKeys]::Sendwait("%{F4}")`

Running tests on my box I get this. 

```powershell
    Context Setting the background to highlight different rows, use of grid view.
      [-] Error occurred in Context block 5.48s
        ParameterBindingValidationException: Cannot bind argument to parameter 'ExcelPackage' because it is null.
        at <ScriptBlock>, C:\Users\Douglas\Documents\GitHub\ImportExcel\__tests__\Compare-WorkSheet.tests.ps1: line 72
        at Invoke-Blocks, C:\Program Files\WindowsPowerShell\Modules\Pester\4.1.1\Functions\SetupTeardown.ps1: line 140
        at Invoke-TestGroupTeardownBlocks, C:\Program Files\WindowsPowerShell\Modules\Pester\4.1.1\Functions\SetupTeardo
wn.ps1: line 130
        at DescribeImpl, C:\Program Files\WindowsPowerShell\Modules\Pester\4.1.1\Functions\Describe.ps1: line 166
```

jhoneill:
CompareWorksheet.Tests doesn't force load the module from the right path, if it finds the previous release that's the error I get :-)

PR #408, opened 2018-08-01
Argument completers for Number format and Worksheetname, bug fix for hidden in set format & more

jhoneill:
- Fixed a bug (typo) in Set-Format which caused -Hidden not to work 
- Import-Excel now has an argument completer for Worksheet name - this can be slow on large files
- The NumberFormat parameter (in Export-Excel, Set-Row, Set-Column, Set-Format and Add-ConditionalFormat) and X&YAxisNumberFormat paramaters (in New-ExcelChartDefinition and Add-ExcelChart) now have an argument completer and the names Currency, Number, Percentage, Scientific, Fraction, Short Date ,Short time,Long time, Date-Time and Text will be converted to the correct Excel formatting strings. 
- Added new function Select-Worksheet to make a named sheet active: Added -Activate switch to Add-Worksheet, to make current sheet active, Export-Excel and Add-PivotTable support -Activate and pass it to Add-Worksheet, and New-PivotTableDefinition allows it to be part of the Pivot TableDefinition. 
- Set-Row and Set-Column will now create hyperlinks and insert dates correctly
- Additional tests. 

I've got one investigate left on my to do list, and all the other things are either improving help or examples/Tests (which as I like to say are all really the same thing).

### DIFF
--- a/AddConditionalFormatting.ps1
+++ b/AddConditionalFormatting.ps1
@@ -122,7 +122,7 @@
         $rule.Formula2 = $ConditionValue2
     }
 
-    if ($NumberFormat)      {$rule.Style.NumberFormat.Format        = $NumberFormat      }
+    if ($NumberFormat)      {$rule.Style.NumberFormat.Format        = (Expand-NumberFormat  $NumberFormat)      }
     if ($Underline)         {$rule.Style.Font.Underline             = [OfficeOpenXml.Style.ExcelUnderLineType]::Single }
     if ($Bold)              {$rule.Style.Font.Bold                  = $true              }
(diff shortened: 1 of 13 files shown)


dfinke:
I saw that when I used `-IncludePivotTable` in the latest release it did not make the Pivot Table sheet the focus, I'm guessing the `-Activate` switch can do that? I'll play with it.

jhoneill:
Yes. The issue was if the there was an existing sheet then an extra sheet didn't become active, but you if you added in some circumstances you could get two sheets pages selected (because it didn't unselect the others).

PR #411, opened 2018-08-02
Fixed Parameter issues in Add-ConditionalFormat & test for it 

jhoneill:
Already fixed issue with 0 values, 0 Enums and specifying -Switch:$false to set format. 
Same issue affected Add-ConditionalFormat ; - this would crash previously if _n_Iconset was specified with the set being "arrows" which a enum item with a zero value.

Resolved this and added a test for it.


dfinke:
Thanks for the updates!

Issue #415, opened 2018-08-09
NoTotalsInPivot Does not disable grand totals


ACiDGRiM:
Hello, I'm using 5.1.1 and am trying to create a report that outputs totals for various status values into a report.
The pivot table is perfectly readable, however the Grand Totals is confusing because the data should be only considered in the context of each pivot column, rather than as a whole set.

It seems the NoTotalsInPivot switch doesn't actually disable the grand total column

jhoneill:
Yes it only removes the grand totals row at the bottom and _not_ the grand totals column on the right … I'll look into this a bit further.

dfinke:
Agreed. So the dll doesn't support column only?

jhoneill:
The DLL has separate properties for Row and column (or colum as it writes it). What we have at the moment only sets the Row one. It seems you have all for combinations of true/false on each. Which will be in the next PR :-)

jhoneill:
As of the newest merge we have `-PivotTotals` which can be rows, columns both or none. We still have NoTotalsInPivot which is equivalent to` -PivotTotals None`

Issue #409, opened 2018-08-01
[OfficeOpenXml.ConditionalFormatting.ExcelConditionalFormattingCollection]


dfinke:
Version 5.1.0 published on the gallery has this issue, the previous one is good.

[OfficeOpenXml.ConditionalFormatting.ExcelConditionalFormattingCollection] does not contain a method named 'Add'.

![image](https://user-images.githubusercontent.com/67258/43551618-9fdc1746-95b5-11e8-946a-2d1757f2e069.png)

jhoneill:
https://jamesone111.wordpress.com/2018/07/30/on-powershell-parameters/ 
Strikes again. 

I refactored things to call Add-ConditionalFomatting. which is the change which causes it to go   

It gets knocked into a cocked hat by enums which have an item with a zero value  - this bug has been in Add-conditionalFormatting from the outset :-( 


So I've done the same it as I did to set format, and I've added your example to the tests  :-)  Just running a test pass and then I'll commit and create a PR.

jhoneill:
Fix with the merge of 416

PR #416, opened 2018-08-12
Fix for #415, auto extend ranges/tables on append, password fixes, wildcards in Exclude property

jhoneill:


dfinke:
Thanks @jhoneill, I'll take a look. I was playing with Import-Excel, two cols in the sheet, raw data, no headings, name and age. `Import-Excel -noheader` works assigning p1 and p2 as property names, using `-HeaderNames name, age`, it does it mostly right, but skips the first data row. I took a quick look, not obvious yet to me, wondering if you had thought off the top of your head? I was the last one in there though.

jhoneill:
IIRC, -noHeader says "there is no title row", and -HeaderNames says "Use these names instead"  I think for the case where you have no titles and don't want P1, P2 it needs both.

dfinke:
Yeah, tried that, and got parameter issues. Will take a deeper look

dfinke:
I think the `-NoHeader` means there are no headers, property names to be extracted. 

So this line causes the off by one error. I took out the `1 +`, got the correct results. Need to work up tests. 
https://github.com/dfinke/ImportExcel/blob/master/ImportExcel.psm1#L381

The parameters on this are a bit mystifying. IIRC the start, bottom, left, right would let your target a range of cells that may or may not have property names. It could also be used to bypass a title row.

jhoneill:
We should have start row - ignore rows above it Then we have four permutations, 
Start row is headers - use them
Start row is headers - but use something else
Start row is data - name columns P1, P2 etc
Start row is data - use something else. 

If you take out the 1 + it will fix one scenario and break another :-(

dfinke:
Thanks, glad I asked. I'll let it marinate and then move forward.

dfinke:
Curious, did you use `\d` over `[0-9]` for perf?

jhoneill:
I haven't tested the perf of [0-9] vs \d  ; I'd expect the latter to be faster but the margin should be tiny. 

Using `[regex]` syntax should be faster than `-match`. But the main thing is the that 
`$x -match "\d"`  does an implicit `$x.tostring()` and 
`$regex.ismatch() `does an implicit `[String]$x 
`
You think they were the same … but try 
```
$s = get-service -Name LanmanWorkstation
[string]$s
$s.tostring()
```

dfinke:
I was guessing it's be along those lines. Another blog post? :)

dfinke:
@jhoneill I'll merge this if you're good to go

jhoneill:
Yes Merge it now or it will go on forever !

jhoneill:
#415, #419 and #409 can now be closed :-)

Issue #430, opened 2018-09-05
Question: Is it possible to add conditional formatting using a ThreeColorScale in reverse order?


danstis:
I am looking to add conditional formatting, using a ThreeColorScale but have the low numbers colored green and the high colors in red. Is there any way to achieve this?

Thanks, Dan

dfinke:
Please could you post the script and xlsx, thanks

jhoneill:
Yes. It's possible. You have to use `Add-ConditionalFormat` with the `-PassThru `switch which was new in 5.2 
```
$excel = dir | select name,length,LastWriteTime |export-excel -path "$env:temp\test99.xlsx" -AutoNameRange -PassThru
$rule = Add-ConditionalFormatting -Address $excel.Workbook.Worksheets[1].cells["length"] -RuleType ThreeColorScale  -Passthru
$rule.LowValue.Color = [System.Drawing.Color]::LightGreen
$rule.MiddleValue.Color = [System.Drawing.Color]::LightYellow
$rule.MiddleValue.Type = "Percentile"
$rule.HighValue.Color  = [System.Drawing.Color]::OrangeRed
Close-ExcelPackage $excel -Show
```
By default the middle value is 50 Percent of the maximum but you can change the `.type `(possible values are Formula, Max, Min, Num, Percent, Percentile default is Percent) and `.value` (default is 50) intellisense will complete` [System.Drawing.color]::` for you with known colours

dfinke:
@jhoneill Bummer, I don't see a test for Add-ConditionalFormatting. Did I miss it? I can add a ToDo.

jhoneill:
The tests for add-conditionalformatting  are in export-excel.tests.ps1. I've added some more in my latest commit (PR soon!) .

danstis:
Thanks @jhoneill this explains it a bit better.

Still having issues applying it only to the data section of a table, but that is not directly related to this question, I'll figure that one out.

jhoneill:
I find that I get one of two problems. Either the rule matches the header and formats it, or I format the whole column.  What I have taken to doing is using `-AutoNameRange` to give me a ready made name for each block of cells and I use that - which I had in the example. 
 You can also get the address of the table and specify the cells[  ] part using that. If you get stuck, let me know and I'll try to post and example.

jhoneill:
[...]

Issue #426, opened 2018-08-23
Question: How to generate hyperlinks to worksheets on same workbook


heumann-a:
Is there a way to set hyperlinks in cells to other Worksheets in the same excel file?

We want to build something like a "Homepage" with Quicklinks to specific places in the workbook instead of searching every time the whole document.

jhoneill:
Yes.  One of the recent versions added processing of hyperlinks. 
If you have text in this format : 
`"xl://internal/sheet1!A1"`
Export-Excel will process it as a hyperlink - if you link to cell A1 it will say "sheet1" as the text, anything else will say "sheet1!Z10" or whatever.
 
I haven't tested linking to a name from this code but I think you can write "`"xl://internal/sheet1!A1"`

If you look at Export-Excel.ps1 , at line 516 you can see how I construct the link and put it in the cell and set the format so it looks like a link, if you have a list of display text and destinations you should be able to re-use that code.

dfinke:
Started to experiment with this. Need to think through how line 516 might be extended so inter-workbook links can handled.

```powershell
$xlFile = "$env:TEMP\testLink.xlsx"
rm $xlFile -ErrorAction Ignore

$map = @{
    A2="sheet2!A1"
    A3="sheet3!A1"
    A4="sheet4!A1"
}

@"
Links
Goto Sheet2
Goto Sheet3
Goto Sheet4
"@ | ConvertFrom-Csv | Export-Excel $xlFile -AutoSize 

"Text`r`nSheet2" | ConvertFrom-Csv | Export-Excel $xlFile -AutoSize -WorksheetName Sheet2
"Text`r`nSheet3" | ConvertFrom-Csv | Export-Excel $xlFile -AutoSize -WorksheetName Sheet3
$pkg = "Text`r`nSheet4" | ConvertFrom-Csv | Export-Excel $xlFile -AutoSize -WorksheetName Sheet4 -PassThru

$ws = $pkg.Workbook.WorkSheets["Sheet1"]

$map.GetEnumerator() | % {
 
    $TargetCell = $ws.Cells[$_.Key]
    $display = $TargetCell.Text
    $referenceAddress = $_.Value

    $h = New-Object -TypeName OfficeOpenXml.ExcelHyperLink -ArgumentList $referenceAddress, $display
    $TargetCell.HyperLink = $h
    $TargetCell.Style.Font.Color.SetColor([System.Drawing.Color]::Blue)
[...]

Issue #424, opened 2018-08-22
Conditional formatting on a table produces a currupted xlsx


Artanor:
I am trying to add some conditional formatting to a table but it results in a corrupted file.

Using the  MultipleTables.ps1 example if I modify the last table like so.
````
$ps |
Sort-Object PM -Descending |
Select-Object -First 10 company, PM |
Export-Excel $xlfile -WorkSheetname Handles -AutoSize -TableName PM -StartRow 13 -ConditionalText $(
	New-ConditionalText -ConditionalType Equal "Microsoft Corporation" -Range "A:A" -ConditionalTextColor Black -BackgroundColor Coral
	New-ConditionalText -ConditionalType GreaterThan 400 -Range "B:B" -ConditionalTextColor Black -BackgroundColor Coral
)
````
The resultant xlsx has an error that mentions a corrupted Sheet2.xml

jhoneill:
Hi 
`Equal "Microsoft Corporation"`

Needs to be `
Equal '"Microsoft Corporation"'`
I think this applies to any string with the greater than / less than / equals operators. (When I looked in Excel it always puts quotes around strings with these. I'll have a look and see if we can make things more robust.

PR #435, opened 2018-09-11
Improved help, minor bugfixes and parameter improvments, more tests

jhoneill:
-PassThru is available in more places. 
-Renamed Set-Format -Row -Column to have Excel in the name (old names as an alias). 
-Set-Format and Add-ConditionalFormat now use "Range" and "Address" in line with EPPlus, and are easier to provide parameters to (backwards compatibility maintained).  Add-ConditionalFormat supports reverse for DataBars (#430)  and wraps strings in quotes when they need it and aren't wrapped (#424)  
-Better handling of links (#426) 
- Move Add named range and Add table functionality out of Export Excel and into separate functions Add-Table supports additional table options (e.g. totals). 
- Add-PivotTable is its own file and now calls Add-ExcelChart to create Pivot charts, and supports a chart definition as a parameter (as well as Address [to locate the pivot *not* on a new page] and PivotTableStyle) 
- ConditionalFormatting, Chart, and Pivot definitions can all be passed as hash tables (the original objects are still supported).   
- Help and example improvements.


DarkLite1:
Great PR, just one question about the help. It always seems hard to find the help because `Get-Help ...` doesn't always show everything. And I noticed that when you run `Install-Module ExportExcel` you will not get the folder `Examples` (Examples/PivotTable/TableAndPivotTable.ps1). I don't know if this is by design..

jhoneill:
@DarkLite1  I've tried to fix the thing with the help, basically comment based help can be too sensitive to the order things appear in the file, so what looks like perfectly good help gets ignored. 
I wrote myself a tool which I'll share at some stage - which checks that the for each command in a module there is a description, a synopsis, counts the examples and checks there is parameter help for everything which isn't a common parameter, and I've fixed up the bits that I've worked on. 
The examples (and tests) @dfinke  can speak to better, but there is a trade off between having a small download and having all the examples and tests.

dfinke:
@jhoneill is there a Script Analyzer rule that can be enabled for help?

When I started building the Examples, I wanted folks to visit the repo to check things out. The repo examples were what I could point to for "How do I questions..."

DarkLite1:
I actually never really knew there was an Example folder in the first place. Always depended on `Get-Help` to tell me what I need. Anyhow, I believe it deserves more spotlight or should be included in the regular help for those specific CmdLets. #JustMyTwoCents :)

jhoneill:
@DarkLite1  - there's a size where something becomes too big for a help example, but as a general rule, yes, help is the best place, and should be the default things should only go elsewhere if they are too big and unwieldy for on-line help. 

@dfinke  I don't think Script analyser does it (Otherwise a it would go berserk when a lot of things were brought into code). I might have to polish up and share my script :-)

dfinke:
Hmm, maybe in the inline help add links to the examples on GitHub? :)

DarkLite1:
@dfinke that would be the best I think. Just use `.LINK` and point it to the right folder for that CmdLet. I assume each CmdLet has its own folder? Then it would be inline and easy to find. 

Maybe a remark on the home page would be good to, how to find the folder `Example` with some demo's.

dfinke:
@DarkLite1 Yup, be great to get some help on that too ;-)

DarkLite1:
Very busy currently with a big code migration. But when I find some time I'll try to help out 👍

ili101:
I think that if you want to make this the most user friendly the best thing to do is to use the GitHub Wiki.
Create a .md page for each CommandLet in the Wiki with basic explanation and an example or a link to the already created example files, redirect the online help of the CommandLets like "Get-Help Export-Excel -Online" to the appropriate Wiki page.
 
Also README.md got way to long, probably best to split it to a [CHANGELOG.md](https://github.com/PowerShell/PowerShell/blob/master/CHANGELOG.md) file and Wiki pages.

Only downside it will probably take some time to set up :)

PR #436, opened 2018-09-17
Fix AutoNameRange Parameter to replace spaces with a delimiter

dafyre:
This PR fixes AutoNameRange to replace Spaces in Name Range with a delimiter.

Added -AutoNameDelimiter parameter.  It defaults to _


dfinke:
Needed to merge an earlier branch, that's why there are conflicts. I like the idea.

I'm wondering, if I have a property name "Product Sales" and use `AutoNameRange` and under the covers is silently changes the name and then in `PivotColumns` I use it, it won't work and probably be difficult to debug.

Two things I'd prefer `AutoNameRangeDelimiter` for the name and need some unit tests.

dafyre:
You actually have a point about PivotColumns.  Admittedly, I am not too familiar with using Pivots in Excel.  The same code would need to be updated to use the AutoNameRangeDelimiter there as well.
  
Are spaces actually allowed for the named ranges?  This is what got me trying to add this in.  I have some data with spaces in the column headers and data, like:

**Date** | **Product ID** | **Product Sales**
9/10/18  |  123456           |  10487
9/11/18  |  123456           |   9762
9/12/18  |  123456           |   12627

Export-Excel generates the Excel File, but Excel 2016 will not open it correctly.  It will show my data, but the chart that I have added to the excel sheet using new-excelchart shows up blank, and excel fusses about things not being right.  It only happens on documents where my column headers have spaces.

dfinke:
Thanks. There are merge conflicts and test failures. Need that reviewed. 

Thoughts on questions I raised?

dafyre:
> Thanks. There are merge conflicts and test failures. Need that reviewed.
> 
> Thoughts on questions I raised?

Thoughts about the PivotColumns?  I don't know right off, I'd have to go experiment.

As far as naming the variable.  I changed it as you suggested to AutoNameRangeDelimiter.   I just pushed that change in.  and adjusted line 773 (``$targetRangeName = $script:Header[$c] -replace '\W' , $AutoNameRangeDelimiter``) I'm trying to figure out why the merge fails, but don't see anything glaringly obvious.

dfinke:
Yeah, I definitely like the idea. Magically changing data our from underneath a user is problematic, and this may just be shining the light on solving a more general issue.

The named ranges can be used in many places. I think it could cause a lot of debug pain if property names are changed implicitly. Not sure how to solve it yet,

@jhoneill has tackled this in several areas.

dafyre:
> Yeah, I definitely like the idea. **Magically changing data our from underneath a user is problematic**, and this may just be shining the light on solving a more general issue.
> 
> The named ranges can be used in many places. I think it could cause a lot of debug pain if property names are changed implicitly. Not sure how to solve it yet,
> 
> @jhoneill has tackled this in several areas.

No kidding, lol.  I took a bit of a closer look when looking to see where else $AutoNameRangeDelimter would need to go, and I saw warnings being displayed where things weren't right.  

As long as the user knows what you are magically doing to the data, it makes it possible for them to (potentially easily) work out a plan.  For instance, on said chart that I was having a problem with, I am able to account for the underscore in my field names...  

~~~
        $dlpFile="DLPPolicyMatches_$reportSuffix"
        $dlpHeader="Severity Low","Severity High"
        $dlpSeries=$dlpHeader -Replace(" ","_")
~~~
I can replace the spaces in my column names with _'s since I know that's what the Export-Excel module does.

Perhaps it might be a good idea to make this a universal delimiter used anywhere that your code tries to 'magically change data from underneath a user'.  Call it -exportDelimiter or something.

jhoneill:
So … the version I have checked in has this line
`
$targetRangeName = $script:Header[$c] -replace '\W' , '_'`

So anything that is not a "word character" will become _ in the when the range name is created. Having a parameter to set what is used as the replacement is a good refinement. 


It doesn't matter for Pivot columns because that is going to use what appears at the top of the column
So if my data is defined starting 
```
$data = ConvertFrom-Csv @"
Date,Buyer name,Type,Amount
```

and I do 
`$data | export-excel -Path foo.xlsx -AutoNameRange -IncludePivotTable -PivotRows "Buyer name" -PivotColumns "Type" -PivotData @{"Amount"="Sum"} -Activate -Show`

The range for the second column is 'Buyer_Name' but the field in the pivot table is "Buyer Name" with the space, so Pivots are OK, but there is still a problem. 
If I do
```
$excel = $data | export-excel -Path foo.xlsx -AutoNameRange -IncludePivotTable -PivotRows "Buyer name" -PivotColumns "Type" -PivotData @{"Amount"="Sum"} -Activate -PassThru
Set-ExcelRange -Range $excel.Sheet1.cells['Date'] -NumberFormat 'Short Date'

```
 I can use the named range to apply formatting (cool huh ? Only works with the newest version. $excel.SheetName is also new ) 
But now if I try to use  `Set-ExcelRange -Range $excel.Sheet1.cells['Buyer Name'] -bold
` there is no such range and the error message is not helpful. I have to put the _ in. 

Now … if you try to add a range like this `Add-ExcelName -Range $excel.Sheet1.cells['a3:D3'] -RangeName "how about this"`  that generates a warning `Range name 'how about this' contains illegal characters, they will be replaced with '_'.`  But what I've done in Export-Excel circumvents the warning.  I wonder if it would be better to take that line out of Export-Excel and let Add-ExcelName fix up bad names and generate warnings as it goes. (Which also means passing it the replacement character parameter) At the moment the user is left wondering why  `$excel.Sheet1.cells['Buyer Name']` says it doesn't have a bold property but `$excel.Sheet1.cells['Date'] ` does … that kind of magic creates confusion. A warning which says, "Hey I didn't like the name 'Buyer Name'"  would give them a clue. 

Thoughts ?

dfinke:
Yeah I like this.

Automatic global translation for ranges makes sense. That is cool. I vaguely remember hitting issues with casing of names, but that's off topic.

When I'm knocking out a an ImportExcel script these were the "kind of" under the covers type things that took me down the rabbit hole for too long.

dafyre:
What about handling this somewhere that everything passes through?

I've taken Add-Excel, and Set-ExcelRange and done this. I went through everywhere that had ```$SourceRange -is [string]``` I added a RangeDelimiter parameter to each one with a default of underscore (for this module, the underscore seems to be a sensible default).

Take a quick glance and tell me if you think this is the right approach or not.  (Appveyor doesn't like something I did.  I'll kick it around in a bit.

dfinke:
Thanks for walking that through @dafyre. Yes, centralizing this is key, going to take a look at it in the overall scheme and see what makes sense.

dfinke:
Will do. Thanks for the PR

PR #444, opened 2018-09-27
Minor improvements

jhoneill:
Small bug fix on -AutoNameRange and now allow 
`1..5 | set-column -worksheet $ws -AutonameRange
`
Test for an empty dataset in send-sqlDataToExcel didn't do the job . 
JoinWorksheet didn't have to color argument completer and couldn't take background color on its own. 
Better tests, examples and Help.

### DIFF
--- a/AddConditionalFormatting.ps1
+++ b/AddConditionalFormatting.ps1
@@ -1,7 +1,7 @@
 ﻿Function Add-ConditionalFormatting {
     <#
       .Synopsis
-        Adds contitional formatting to worksheet.
+        Adds conditional formatting to worksheet.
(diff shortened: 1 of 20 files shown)


jhoneill:
You wouldn't believe how many times "postion" appeared instead of "position". I dumped a lot of help into Words and fixed most of the things words saw as wrong, but it needs a proof reader to go through it properly. Fortunately, I'm married to one :-)

dfinke:
Lol, really. I was thinking how to automate this. But grabbing help based text is a challenge. I use spellright for markdown in vs code.

PR #446, opened 2018-10-02
Expand a few aliases

lazywinadmin:


dfinke:
Thanks @lazywinadmin !

PR #458, opened 2018-10-08
Fixed #457 & a couple more tweaks

### LINKED ISSUE #457: Set a custom font name and size.
Hi there. In the cmdlets `Set-ExcelRow`, `Set-ExcelColumn` and `Set-ExcelRange` the following parameters are not called anywhere in the functions.
```
[String]$FontName,
#Point size for the text.
[float]$FontSize,
#Change background color.
```
Is there a way to set a custom font name and size? 
Thank you.

jhoneill:
Found Import-Excel had Write-verbose for each cell; removed this for perf. 
Some parameter combinations with Merge-MultipleWorksheets didn't work as they should, because of an issue in Merge-Worksheet, so fixed that. Also fixed a conditional format bug in Merge-Worksheet which meant some backgroundColor didn't show because the background pattern was none instead of solid. 
Code for setting font name was missing. Added to it set-excelRange and made sure it was passed from set excel-Row/column. Added teststs for these.

### DIFF
--- a/ImportExcel.psm1
+++ b/ImportExcel.psm1
@@ -327,11 +327,11 @@ function Import-Excel {
             $Path = (Resolve-Path $Path).ProviderPath
             Write-Verbose "Import Excel workbook '$Path' with worksheet '$Worksheetname'"
             $Stream = New-Object -TypeName System.IO.FileStream -ArgumentList $Path, 'Open', 'Read', 'ReadWrite'
-        } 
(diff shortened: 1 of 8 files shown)


dfinke:
Thanks for the fixes on #457.

I'll work to publish this soon. If you need it now, you can grab the zip here.

https://ci.appveyor.com/project/dfinke/importexcel/builds/19353077/artifacts

PR #474, opened 2018-10-23
Create AddMultiWorkSheet.ps1

conradagramont:
This example shows how to add multiple tabs within the same file.


dfinke:
@conradagramont Thanks for the example. This can be simplified. You don't need to do the `-PassThru` and  use `Close-ExcelPackage`. Just use ` Export-Excel` and on the last one use the `-Show`. If you can re-work that and commit it, should be able to merge it then.

conradagramont:
Thanks for the for the feedback Doug!  I checked in the change.  It's cool to see that contributing my example actually helped me improve my own code!  Community coding at it's finest!!

dfinke:
Great! Do you have a Twitter handle I can use in the comments?

conradagramont:
Twitter: @agramont

PR #475, opened 2018-10-24
Colors for PS core and proof reading

jhoneill:


jhoneill:
I didn't really look at what was being returned by the V6 CI. I know that there are some errors and warning which really should be investigated, but time has been a bit short :-(

dfinke:
Cool. Yeah, I turned on just to see. Happened to notice the color string issue. Waiting till epplus is out of preview

jhoneill:
Just seeing appveyor doing the tests on 6.1 and classing it as a fail.  That's not really desirable until we get to fully supporting 6.x

dfinke:
Yup, just thought it was a bit useful to see the color string issues

dfinke:
@jhoneill, move ImportExcel to v6.0?

jhoneill:
I'd probably hold off changing the major version until it's based on the current EPPlus and we feel confident with x-plat V6 support. 
I've just fixed a couple of things in the tests so that V6 passes (A V6 process object has more parameters including a code property, Parent, which return a process which doesn't export, so a little handling in the tests fixes that). I'll do another PR for those

dfinke:
Sounds good

Issue #479, opened 2018-10-28
New-ExcelChart -LegendPosition top


jotaene:
Hi

This is an extraordinary module, thanks for your work!

It seems that with -LegendPostion 'Top' I always get 'Right' position of legend.

Thanks again.

dfinke:
Thanks! Could you please post and example of the script, xlsx, and version of the module.

jotaene:
Yes.

ImportExcel Version: 5.3.4
xlsx: [Test.xlsx](https://github.com/dfinke/ImportExcel/files/2523169/Test.xlsx)
script: [script.txt](https://github.com/dfinke/ImportExcel/files/2523171/script.txt)

Thank you

dfinke:
Looks like a bug. 

https://github.com/dfinke/ImportExcel/blob/master/New-ExcelChart.ps1#L404

`$LegendPosition` does not evaluate to `$true` when specified. Perhaps cause it is an Enum.

# Hacked it 

This works, but does not work when `$LegendPosition` is not specified or defaulted with "Right"

`if ($LegendPosition.ToString().Length -gt 0) { $Chart.Legend.Position    = $LegendPosition }`

Need to think about it. @jhoneill my first time hitting this with an enum, not sure that is the root cause though.

jotaene:
Now it works. Thanks again.

jhoneill:
Yes. It is the same bug I've introduced before 
`If ($foo) {do something} `
Fails when `$foo` is zero. I thought I'd flushed most of them out 

 I'll take a small bet that  `[OfficeOpenXml.Drawing.Chart.eLegendPosition]::top ` is zero. 
Your fix will work fine, elsewhere we've used 
`if ($PSBoundParameters.ContainsKey('LegendPosition'))`

At some point I will have to write something which gets the parameters for each function and looks to see if it contains if ($parameter)

jotaene:
Done and working like a charm. Thanks for your attention.

PR #482, opened 2018-10-29
Parameter clean up, improvements to Merge-MultipleSheets

jhoneill:
#479 found another case of `If ($parameter) {do stuff with parameter} ` failing when `[Boolean]$parameter` returns false. This can be when -Switch:$false should be allowed; when `$parameter ` is an enum where zero is a valid value; or is string where an empty string is allowed, or a number where 0 should be allowed. 
in these cases in is preferable to use
`if ($PSBoundParameters.ContainsKey('Parameter')) `
It is OK to use the other format when zero, null or empty are not allowed, or for a switch where $false is never used. 
This includes a checkin after a search for that being inappropriately, and changes to fix a couple of small issues with Merge-Multiplesheets.


jhoneill:
Yes the $ is misplaced. Good catch! 

(note to self, test adding a "between" conditional format - that got through because there isn't one).

dfinke:
Thanks!

PR #484, opened 2018-11-01
fix for loading on case-sensitive filesystems

rfdrake:
On linux/docker, powershell can't load some of the modules.

Here is an example of the error message:

    . : The module '/usr/local/share/powershell/Modules/ImportExcel/5.3.4'
    could not be loaded. For more information, run 'Import-Module
    /usr/local/share/powershell/Modules/ImportExcel/5.3.4'.
    At
    /usr/local/share/powershell/Modules/ImportExcel/5.3.4/ImportExcel.psm1:7
    char:7
    +     . $PSScriptRoot\Compare-WorkSheet.ps1
    +       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound:
    (/usr/local/shar...e-WorkSheet.ps1:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CouldNotAutoLoadModule

The reason for this is that the filenames on disk have different
capitalization than the filenames in ImportExcel.psm1.

### DIFF
--- a/ImportExcel.psm1
+++ b/ImportExcel.psm1
@@ -4,7 +4,7 @@
     . $PSScriptRoot\Charting.ps1
     . $PSScriptRoot\ColorCompletion.ps1
     . $PSScriptRoot\ConvertExcelToImageFile.ps1
-    . $PSScriptRoot\Compare-WorkSheet.ps1
+    . $PSScriptRoot\compare-worksheet.ps1
     . $PSScriptRoot\ConvertFromExcelData.ps1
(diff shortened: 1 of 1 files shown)


dfinke:
Thanks!!

I saw that the next preview of PS Core is going to help/remove case-sensitivity on non-Windows platforms.

Thanks for updates, this module is going to go through some growing pains in this space.

costleya:
Can we merge this in? It looks like an uncontroversial change.

PR #523, opened 2019-01-12
Update Export-Excel.ps1

stahler:
Love this module!  I evangelize it often!  Extremely minor adjustment in the help.  Awesome work!


dfinke:
Thanks for the kind words! Thanks for taking the time to do the pull request.

PR #527, opened 2019-01-17
fixed lastColumn issues when more than 10 rows exist in excel sheet

JakobGSvendsen:
I had an issue where the dimension value would be
E17
So the old regex would output E1 as last column, which made the other commands fail.

Updating the regex to only accept letters as 0-9 is part of \w

### DIFF
--- a/Export-Excel.ps1
+++ b/Export-Excel.ps1
@@ -1036,7 +1036,7 @@
         if ($CellStyleSB) {
             try {
                 $TotalRows = $ws.Dimension.Rows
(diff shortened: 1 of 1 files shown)


dfinke:
Thanks for the PR. I'd prefer if there were tests that proved the failure. Then, adding this code, they would pass.

JakobGSvendsen:
alright, makes sense. It is a little difficult when the erorr is inside a scriptblock that is used internally in the command, but I guess the best way is to just save LastColumn in script or global scope and check it . sounds like a good task for me in pester. Will update asap hopefully today

PR #537, opened 2019-01-31
Convert scientific notation to decimal

pwoubshet:
Excel converts very small and very large numbers to scientific notation which breaks imports to SQL Server columns expecting a numeric format.  This change will convert it from scientific notation back to decimal.


dfinke:
Thanks! Could you please add a test for this, it'll streamline merging it.

tomvanstraelen:
Hello,

When importing an Excel file with a small number, I still get the value in scientific notation.
This is with version 7.1.0
Is the fix not yet integrated in the latest version ?

Thank you for looking into this.

PR #552, opened 2019-02-25
Minor fixes and added data-validation support

jhoneill:


dfinke:
@jhoneill for the behavior changes, seems like they are breaking, the module is at v5.4.4. EPPlus is not fully ported, so bump it to v5.5.0?

jhoneill:
They shouldn't break anything which is live - the one thing which I found was different - and impacted the tests - was if I had one column and row 1 was null the spreadsheet was only considered to start at row 2. 

The rule used to go 

1. If it's a date
2. If it's a timespan
3. If it's a value
4. If it starts with = 
5  if it look likes a URI 
6  otherwise can we make it a number ? No ? Insert a string 

Then between 3 & 4 I put is it of type URI, and  is it not of type string  - that tried to call the toString method if it was null. Which caused and error. So before 1 I put "is it null, if so do nothing" . 
And I've changed 6
Does it have digits ?  does no-number-conversion apply ? OK now try to make it a number.

jhoneill:
I've put in a fix for #555 as well :-)

PR #554, opened 2019-02-27
Correct type casting syntax (fixes #551).

### LINKED ISSUE #551: Conditional Test Hex Colours stopped working after version 5.4.0. it was working until 5.3.4
Conditional Text Hex Colours stopped working after version 5.4.0. it was working until 5.3.4

Hello Doug, A big Thanks for this EXCELLENT module. I would like to flag that before the hex colours used to work in versions 5.3.4, it stopped working after that. 

Please see the code below:

` $batchname = 'Maintenance25'
 $text2 = New-ConditionalText 'Completed' -ConditionalTextColor White -BackgroundColor '#00b359' # green back ground with white fonts. Worked in 5.3.4
 $text1 = New-ConditionalText 'Failed' -ConditionalTextColor White -BackgroundColor '#FD625E' #red background with white fonts. Worked in 5.3.4
 

$paramExportExcel = @{
	Path = ".\$($batchname)_CutoverReport.xlsx"
	Show = $false
	AutoSize = $true
	TableName = 'CutoverReport'
	TableStyle = 'Light1'
	FreezeTopRow = $true
	BoldTopRow = $true
	WorkSheetname = 'CutoverReport'
	ConditionalText = $text1, $text2
	IncludePivotTable = $true
	PivotRows = 'Status'
	PivotData = 'Status'
	IncludePivotChart = $true
	ChartType = 'PieExploded3D'
}
    
$results | Export-Excel @paramExportExcel # $results has some data`

Please see the Screenshot attached
2019-02-23 11_42_46-exc-mgtbk7p001

miitchyy:
Type casting had a syntax error such that some lines referred to a static member instead of the parameter being cast. Fixes #551.

### DIFF
--- a/AddConditionalFormatting.ps1
+++ b/AddConditionalFormatting.ps1
@@ -193,7 +193,7 @@
     if (-not $worksheet -or $WorkSheet -isnot [OfficeOpenXml.ExcelWorksheet]) {write-warning "You need to provide a worksheet object." ; return}
     #region create a rule of the right type
     if     ($RuleType -match 'IconSet$') {Write-warning -Message "You cannot configure a Icon-Set rule in this way; please use -$RuleType <SetName>." ; return}
-    if ($PSBoundParameters.ContainsKey("DataBarColor"  )      ) {if ($DataBarColor -is [string]) {$DataBarColor = [System.Drawing.Color]::$DataBarColor }
+    if ($PSBoundParameters.ContainsKey("DataBarColor"  )      ) {if ($DataBarColor -is [string]) {$DataBarColor = [System.Drawing.Color] $DataBarColor }
                                                                      $rule =  $WorkSheet.ConditionalFormatting.AddDatabar(     $Address , $DataBarColor )
     }
     elseif ($PSBoundParameters.ContainsKey("ThreeIconsSet" )      ) {$rule =  $WorkSheet.ConditionalFormatting.AddThreeIconSet($Address , $ThreeIconsSet)}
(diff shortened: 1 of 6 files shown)


jhoneill:
re: #551 There was a fix put in to ensure we worked with the PowerShell core - different .NET different behaviour of  [System.Drawing.Color]. 
Previously the Parameters were all of type  [System.Drawing.Color] , 
Windows PowerShell will cast "Red" to a color, and handles Null. 
PowerShell Core won't cast "Red" and also fails on null, breaking the parameters.  

 So the quick fix was to say It can still accept "Red" but it only tries to covert it at the last moment, if it is a string.  

 [System.Drawing.Color]::Red  works on both versions and if $x holds "red" 
 [System.Drawing.Color]::$x  also works. So that syntax was deliberate, in other places  [System.Drawing.Color] $x  (or  $x -as [System.Drawing.Color]) work on Windows PowerShell and fail on core if $x isn't already a color … I thought I had switched them all over. 
This would allow people to pass either a color name, or a [System.Drawing.Color] object on both versions.  **So working as designed.**  But #551 **picks up that the design is flawed** if the string is the hex value for a colour it would cast before, but it now won't work because  [System.Drawing.Color]::#123456 doesn't return anything meaningful. 

Now ... until this evening I was blissfully unaware of [System.Drawing.ColorTranslator]::FromHtml() 
If it works for both names and hex values, and on both versions of Powershell (and it seems to) then that's a better way. Thanks @miitchyy   At some point I'll modify one of the tests so we check for setting color by name and by hex value.

Issue #572, opened 2019-04-03
Export-Excel broken region markup


scriptingstudio:
#region Save an Excel formula - did not close

Issue #557, opened 2019-03-09
Specify *ToExport in module manifest


powercode:
Make sure 
```
CmdletsToExport = @()
FunctionsToExport = @()
AliasesToExport = @()
```
all are specified in the module manifest. Having them empty or as '*' makes forces powershell to load the module when it tries to discover other commands.
This also allows us to not contaminate the global namespace with helper functions.

Issue #555, opened 2019-02-27
Send-SqlDataToExcel.ps1 


canhsyn:
Hi,

I want to add multiple tables to a single excel sheet and then, filter and edit the format. But there is a problem in the filtering stage. Only the last table is being filtered.
Thank you for your help in this matter.

`Send-SQLDataToExcel -MSSQLSERVER -Connection $serverName -SQL $query16first -Database $DBNameSL -Path $newFilename -WorkSheetname "F" -AutoSize -AutoFilter -StartRow 6 -StartColumn 1
`
`Send-SQLDataToExcel -MSSQLSERVER -Connection $serverName -SQL $query16second -Database $DBNameSL -Path $newFilename -WorkSheetname "F" -AutoSize -AutoFilter -StartRow 6 -StartColumn 6 
`
`Send-SQLDataToExcel -MSSQLSERVER -Connection $serverName -SQL $query16third -Database $DBNameSL -Path $newFilename -WorkSheetname "F" -AutoSize -AutoFilter -StartRow 6 -StartColumn 11 
`
`Send-SQLDataToExcel -MSSQLSERVER -Connection $serverName -SQL $query16fourth -Database $DBNameSL -Path $newFilename -WorkSheetname "F" -AutoSize -AutoFilter -StartRow 6 -StartColumn 16
`
`Send-SQLDataToExcel -MSSQLSERVER -Connection $serverName -SQL $query16fifth -Database $DBNameSL -Path $newFilename -WorkSheetname "F" -AutoSize -AutoFilter -StartRow 6 -StartColumn 21    
`

ili101:
I think this cannot be done in excel, can you do it in Microsoft Excel?
You can use -TableName instead of -AutoFilter

canhsyn:
Thanks.  I'm creating too many excel files using the importexcel module. So i need to do this work with using the script. I tried -TableName but it doesn't work.

ili101:
Sorry I wasn't clear, I didn't mean you do it manually I meant that as far as I know this cannot be done even in Microsoft Excel or any other Excel client as this is not supported by the xlsx format.
What's the problem with -TableName ?

dfinke:
# TableName

I didn't look to see if/how `tablename` is passed from `Send-SQLDataToExcel`.

One challenge with tables is, as you filter them the other tables that are close will also change size because Excel adjusts the visible rows .

```powershell
$xlfile="$env:TEMP\testTable.xlsx"
[...]

Issue #506, opened 2018-12-11
Export-Excel extremely slow on large spreadsheets


worc4021:
Hi,

I've been trying to merge two excel spreadsheets together, the source xlsx files are just over 3MB, but when I call `Export-Excel -Verbose` it seems that is flies through the headers and then stalls. These spreadsheets have about 1700 columns and 300 rows, i.e. their union has slightly more. But these numbers are not unheard-of. Is this expected to scale poorly or am I doing something wrong? It stalls saving only one already:

```
$sheet = Import-Excel -Path \\path\to\my\xlsx
$sheet | Export-Excel \\path\to\my\output -Verbose -Debug
```

I can see that it is trying to do something in the task manager, but it's not progressing after

```
VERBOSE: Cell '1:1737' add header [...] 
```

Cheers,
Manuel

worc4021:
Hi! 
Thank you for your quick response. Unfortunately, the reason I left out most of the useful bits from the verbose output is that the spreadsheets I am trying to combine hold simulation output metrics, i.e. work related. So I won't be able to share the spreadsheets, however, the spreadsheets hold very simple data, each row corresponds to one simulation, i.e. each data field is either empty, a string or a numerical value - no functions or anything like that.
I hope this still helps.
Manuel

dfinke:
Ok, if you only do an import-Excel, does it work? What's the response time like?

So 1700x300 empty, string and numeric data?

worc4021:
Import-Excel is fine, takes about 1-2 seconds. 1700x300 was the one I tried, naturally since each row corresponds to one sim and each column corresponds to one metric the size can vary, but that's the ballpark.

dfinke:
I created a sample set of objects. Took 22 minutes to export to excel. There may be some low hanging perf improvements that can be done, not sure how much effort or how much improvement it would bring.

worc4021:
Thank you for looking into this!
- So it is nothing I'm doing particularly wrong on my side. 
- Not sure whether this would be worthwhile.
I just stumbled upon ImportExcel and thought it would be a great prost processing tool for when I compile my regressions to compare different sets of sims. The alternative is literally click 'Add rows' in Spotfire instead. So if you don't see something that obviously makes it scale poorly AND know an alternative implementation, I would not advise you to change the existing version.

Sad times.. I prefer scripting over clicking :D

Thank you,
Manuel

pkarunkar:
I had the same issue but seems like happening only when using verbose
[...]

PR #573, opened 2019-04-04
Perf improvements for Export-Excel

jhoneill:
I had to roll back the changes I made to the readme so github would accept the changes could be merged. So here .... 
# What's new this time. 
- **Performance improvement to Export-Excel** see #506 and #555. This has meant taking code in `Add-CellValue` back into process block of Export-Excel, as the overhead of calling the function was a lot greater than time executing the code inside it. [Blog post to follow](https://jamesone111.wordpress.com). Some tests are showing a ~10x speed increase. #572 was about a broken #region tag in this part of the code and that has been cleaned up in the process. 
- Export-Excel now has an `-InputObject` parameter (this was previously `-TargetData` , which is now an alias for InputObject).
 If the `inputobject` is an array, each item will be inserted, so you can run `export-excel -inputobject $x` rather than `$x | Export-Excel`, and if it is a `system.data.datatable` object it will be inserted directly rather than cell-by-cell. Send-SQLDataToExcel takes advantage of this new functionality. There are simple tests for these new items
- Export-Excel previously assumed `-Now` if there were no other parameters, it will now assume `-Now` if there is no `-Path` or `-ExcelPackage`.     
- The .PSD1 file now itemizes the items exported by the module (#557)

### DIFF
--- a/Export-Excel.ps1
+++ b/Export-Excel.ps1
@@ -423,7 +423,7 @@
     [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingPlainTextForPassword", "")]
     Param(
         [Parameter(ParameterSetName = "Default", Position = 0)]
-        [Parameter(ParameterSetName = "Table"  , Position = 0)]
+        [Parameter(Mandatory = $true, ParameterSetName = "Table"  , Position = 0)]
         [String]$Path,
         [Parameter(Mandatory = $true, ParameterSetName = "PackageDefault")]
         [Parameter(Mandatory = $true, ParameterSetName = "PackageTable")]
@@ -747,11 +747,12 @@
                     catch {Write-Warning -Message "Could not insert the '$Name' property at Row $Row, Column $ColumnIndex"}
                     $ColumnIndex += 1
                 }
+                $ColumnIndex -= 1 # column index will be the last column whether isDataTypeValueType was true or false
                 #endregion
             }
         }
         catch {throw "Failed exporting data to worksheet '$WorksheetName' to '$Path': $_" }
-        $ColumnIndex -= 1 # column index will be the last column whether isDataTypeValueType was true or false
+
     }}
 
     end {
--- a/Export-Excel.ps1
+++ b/Export-Excel.ps1
@@ -133,6 +133,8 @@
             Enables the Excel filter on the complete header row, so users can easily sort, filter and/or search the data in the selected column.
         .PARAMETER AutoSize
             Sizes the width of the Excel column to the maximum width needed to display all the containing data in that cell.
+        .PARAMETER MaxAutoSizeRows
+            Autosizing can be time consuming, so this sets a maximum number of rows to look at for the Autosize operation. Default is 1000; If 0 is specified ALL rows will be checked
         .PARAMETER Activate
             If there is already content in the workbook, a new sheet will not be active UNLESS Activate is specified; if a PivotTable is included it will be the active sheet
         .PARAMETER Now
@@ -427,7 +429,8 @@
         [Parameter(Mandatory = $true, ParameterSetName = "PackageTable")]
         [OfficeOpenXml.ExcelPackage]$ExcelPackage,
         [Parameter(ValueFromPipeline = $true)]
-        $TargetData,
+        [Alias('TargetData')]
+        $InputObject,
         [Switch]$Calculate,
         [Switch]$Show,
         [String]$WorksheetName = 'Sheet1',
@@ -453,6 +456,7 @@
         [Switch]$ShowCategory,
         [Switch]$ShowPercent,
         [Switch]$AutoSize,
+        $MaxAutoSizeRows = 1000,
         [Switch]$NoClobber,
         [Switch]$FreezeTopRow,
         [Switch]$FreezeFirstColumn,
@@ -517,7 +521,7 @@
         [Switch]$ReZip
     )
 
-    Begin {
+    begin {
         $numberRegex = [Regex]'\d'
         function Add-CellValue {
             <#
@@ -539,7 +543,7 @@
             )
             #The write-verbose commands have been commented out below - even if verbose is silenced they cause a significiant performance impact and if it's on they will cause a flood of messages.
             Switch ($CellValue) {
-                { $null -eq $_     } { 
+                { $null -eq $_     } {
                     break
                 }
                 { $_ -is [DateTime]} {
@@ -564,16 +568,16 @@
                     break
                 }
                 { $_ -is [uri] } {
-                    $targetCell.HyperLink = $_ 
+                    $targetCell.HyperLink = $_
                     $TargetCell.Style.Font.Color.SetColor([System.Drawing.Color]::Blue)
                     $TargetCell.Style.Font.UnderLine = $true
                     #Write-Verbose  "Cell '$Row`:$ColumnIndex' header '$Name' add value '$($_.AbsoluteUri)' as Hyperlink"
                     break
                 }
                 { $_ -isnot [String]} {
-                    $TargetCell.Value = $_.toString() 
+                    $TargetCell.Value = $_.toString()
                     break
-                } 
+                }
                 #All the remaining options are string as ... only strings get checked for formulas, URIs or being numbers
                 { $_[0] -eq '='} {
                     #region Save an Excel formula - we need = to spot the formula but the EPPLUS won't like it if we include it (Excel doesn't care if is there or not)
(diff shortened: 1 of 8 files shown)


dfinke:
@jhoneill Planning on zipping this with updates to the readme and then socializing the new bits for broader testing. Let me know if you want to still churn on this.

jhoneill:
@dfinke saw the twitter thread. It's now ready to go. I have a new, slim-line version of Send-SQLData which uses dynamicparam{} to get the parameters from Export-Excel, so if things get added or removed it stays in sync. That's a dozen lines of code and trying to be too clever I managed to get it to write changes to the parameters of a loaded function, which caused fun and games with the parameter sets. Last nights commit was an effort to fix that before I understood it properly  when I finished I found things were better without it.  
The SQL thing is the end of the churn for this one - so if you want to socialize now's a good time

dfinke:
Version number 6.0.0 ?

jhoneill:
Ooh major version update, I thought that was being saved for linking up with the latest EPPlus.dll ? 100% your call :-)

dfinke:
Yeah, this is significant. When EPPlus is xplat, then v7.0, it'll, match the upcoming PS v7.0 version :)

DarkLite1:
@jhoneill thanks for the blogpost. Indeed, maybe I need to start thinking more like an IT Pro instead of a computer scientist, because in the end I just want to go home too :) I just didn't know the impact was so high when calling the function a million times instead of having the code in the process block. Another thing learned today, thank you.

jhoneill:
@DarkLite1 - you're welcome. It *is* amazing, 1ms here and 1ms there doesn't matter, until you have a million of them and that's 15 minutes. Putting stuff into nice neat functions is still the way to go when you are only calling them a modest number of times .

DarkLite1:
I was just wondering if a `Scriptblock = {}` would be faster than a `Function` in this specific case. It would still allow to read clean code. But if the performance drops again it's not worth it off course.

jhoneill:
@DarkLite1  It's something I've got on my list of things to test (along with using PS classes). 
Actually I've just thought of a quicker way to test it which I'll do when I exit headless-chicken mode :-)

ili101:
``` PowerShell
function Verb-Noun
{
    $null = $i
}

$ScriptBlock = {$null = $i}

class ClassName
{
    Test($i)
    {
        $null = $i
    }
}
$ClassInstance = [ClassName]::new()

class ClassNameStatic
{
    static Test($i)
    {
        $null = $i
    }
}

$Tests = @{
    Flat              = {$null = $i}
    Function          = {Verb-Noun}
    ScriptBlockDot    = {. $ScriptBlock}
    ScriptBlockAnd    = {& $ScriptBlock}
    ScriptBlockInvoke = {$ScriptBlock.InvokeReturnAsIs()}
    ClassInstance     = {$ClassInstance.Test($i)}
    ClassInStatic     = {[ClassNameStatic]::Test($i)}
    
}

$Script = {
    Measure-Command {
        for ($i = 0; $i -lt 100000; $i++)
        {
            _Script_
        }
    }
}

$Results = @()
foreach ($Test in $Tests.GetEnumerator())
{
    $Result = [Ordered]@{Name = $Test.Name}
    for ($TestNum = 1; $TestNum -lt 5; $TestNum++)
    {
        $ResultTime = [ScriptBlock]::Create(($Script -replace '_Script_', $Test.Value)).InvokeReturnAsIs()
        $Result["Test $TestNum"] = $ResultTime
        Write-Verbose ("Name: {0}, Test {1}: {2}" -f $Test.Name, $TestNum, $ResultTime) -Verbose
    }
    $Result['Average'] = ($Result.GetEnumerator() | Where-Object Name -Like 'Test*').Value | Measure-Object -Property TotalSeconds -Average | Select-Object -ExpandProperty Average
    $Results += [PSCustomObject]$Result
}
$Results | Sort-Object -Property Average | Format-Table
```
Results:
```
Name              Test 1           Test 2           Test 3           Test 4               Average
----              ------           ------           ------           ------               -------
Flat              00:00:00.1288110 00:00:00.1236764 00:00:00.1415012 00:00:00.1390463 0.133258725
ClassInStatic     00:00:00.6844199 00:00:00.6780921 00:00:00.6681650 00:00:00.6642505 0.673731875
ClassInstance     00:00:00.7459588 00:00:00.7391587 00:00:00.7447225 00:00:00.7368168   0.7416642
ScriptBlockDot    00:00:01.5764087 00:00:01.6110786 00:00:01.5886947 00:00:01.5879553 1.591034325
ScriptBlockInvoke 00:00:01.9530320 00:00:01.9028416 00:00:01.9047247 00:00:01.8740416 1.908659975
ScriptBlockAnd    00:00:03.1146379 00:00:03.1860680 00:00:03.1556045 00:00:03.1653391 3.155412375
Function          00:00:04.2551641 00:00:04.2063672 00:00:04.2027835 00:00:04.2381170  4.22560795
```

dfinke:
This might make a nice blog post. PS v5.1? Is PS v6.x better?

ili101:
Yes that was on 5.1. Interesting side note the script actually didn't work on 6.2 I needed to add explicit ToString `$Test.Value.ToString()` where in 5.1 it worked without it. So some work may be required moving scripts to Core. 
So this version works on Core:
``` PowerShell
function Verb-Noun
{
    $null = $i
}

function Verb-NounParam
{
    param
    (
        [Int]$i
    )
    $null = $i
}

function Verb-NounPipe
{
    param
    (
        [Parameter(ValueFromPipeline)]
        [Int]$i
    )
    begin
    {
    }
    process
    {
        $null = $i
    }
    end
    {
    }
}

$ScriptBlock = {$null = $i}

class ClassName
{
    Test($i)
    {
        $null = $i
    }
}
$ClassInstance = [ClassName]::new()

class ClassNameStatic
{
    static Test($i)
    {
        $null = $i
    }
}

$Tests = [Ordered]@{
    Flat              = {$null = $i}
    Function          = {Verb-Noun}
    FunctionParam     = {Verb-NounParam -i $i}
    FunctionPipe      = {$i | Verb-NounPipe}
    ScriptBlockDot    = {. $ScriptBlock}
    ScriptBlockAnd    = {& $ScriptBlock}
    ScriptBlockInvoke = {$ScriptBlock.InvokeReturnAsIs()}
    ClassInstance     = {$ClassInstance.Test($i)}
    ClassStatic       = {[ClassNameStatic]::Test($i)}
}

$Script = {
    Measure-Command {
        for ($i = 0; $i -lt 1000000; $i++)
        {
            _Script_
        }
    }
}

$Results = @()
foreach ($Test in $Tests.GetEnumerator())
{
    $Result = [Ordered]@{Name = $Test.Name}
    for ($TestNum = 1; $TestNum -lt 5; $TestNum++)
    {
        $ResultTime = [ScriptBlock]::Create(($Script.ToString() -replace '_Script_', $Test.Value.ToString())).InvokeReturnAsIs()
        $Result["Test $TestNum"] = $ResultTime
        Write-Verbose ("Name: {0}, Test {1}: {2}" -f $Test.Name, $TestNum, $ResultTime) -Verbose
    }
    $Result['Average'] = ($Result.GetEnumerator() | Where-Object Name -Like 'Test*').Value | Measure-Object -Property TotalSeconds -Average | Select-Object -ExpandProperty Average
    $Results += [PSCustomObject]$Result
}
$Results | Sort-Object -Property Average | Format-Table
```
Results on 6.2:
```
Name              Test 1           Test 2           Test 3           Test 4                Average
----              ------           ------           ------           ------                -------
Flat              00:00:01.3098483 00:00:01.3011428 00:00:01.3415053 00:00:01.3707140    1.3308026
ClassStatic       00:00:06.8755267 00:00:06.9209289 00:00:06.9608326 00:00:06.7384613  6.873937375
ClassInstance     00:00:07.5806316 00:00:07.6416325 00:00:07.5750805 00:00:07.6209510    7.6045739
ScriptBlockInvoke 00:00:10.0704042 00:00:09.6795793 00:00:09.4151863 00:00:09.2425907  9.601940125
ScriptBlockDot    00:00:14.4357740 00:00:14.2959977 00:00:14.3401616 00:00:14.2887840 14.340179325
ScriptBlockAnd    00:00:18.9680652 00:00:18.8571368 00:00:18.8282078 00:00:18.7296331 18.845760725
Function          00:00:27.2381089 00:00:27.1557229 00:00:26.6860846 00:00:26.7743435 26.963564975
FunctionParam     00:00:31.0107950 00:00:30.9357609 00:00:30.9934745 00:00:30.7932976    30.933332
FunctionPipe      00:00:41.6640927 00:00:43.3245141 00:00:41.9259585 00:00:41.4085794 42.080786175
```
Nice improvements on ScriptBlockInvoke, ScriptBlockAnd, Function and FunctionParam

Also this is probably the best case scenario, the impact may be different in Export-Excel but this is a good general starting point.

dfinke:
Could be due to the performance improvements on .NET Core too

PR #576, opened 2019-04-07
Double DataTable performance

ili101:
I was seeing that the new version was twice as slow compared to my old one, after some investigation I found that the ` | Out-Null` on `LoadFromDataTable` was changing it's execution time on the line from 0 to as long as the xlsx saving time doubling the total execution time.

I also replaced all the `Out-Null` in the module just in case it cusses slowdown in other places.
@jhoneill never use `Out-Null` it's pure evil https://stackoverflow.com/a/5263780

### DIFF
--- a/AddDataValidation.ps1
+++ b/AddDataValidation.ps1
@@ -64,7 +64,7 @@
         [String]$NoBlank
     )
     if  ($Range -is [Array])  {
-        [void]$PSBoundParameters.Remove("Range")
+        $null = $PSBoundParameters.Remove("Range")
         $Range | Add-ExcelDataValidationRule @PSBoundParameters
     }
     else {
--- a/ConvertExcelToImageFile.ps1
+++ b/ConvertExcelToImageFile.ps1
@@ -35,15 +35,15 @@
         Write-Progress -Activity "Exporting $range of $workSheetname in $Path" -Status "Opening Workbook and copying data"
         $xlWbk  = $xlApp.Workbooks.Open($Path)
         $xlWbk.Worksheets($workSheetname).Select()
-        $xlWbk.ActiveSheet.Range($range).Select() | Out-Null
-        $xlApp.Selection.Copy()                   | Out-Null
+        $null = $xlWbk.ActiveSheet.Range($range).Select()
+        $null = $xlApp.Selection.Copy()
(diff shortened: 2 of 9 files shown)


jhoneill:
I get the same it roughly halves the time on a large table insert of that line (628) is changed from | out-Null to $null =  
 There's no consistency here, some places we have x | out-Null,  some [void] x ; some $Null = x ;  
Getting rid of the | out-Nulls is a start. Then we should just use one the others, not both :-)

dfinke:
I'm fine with `$null =`. I picked up the `[void]` approach from Jason Shirk at one point.

ili101:
There where only 11 [void]s, mainly in the new test.

jhoneill:
BTW, out-null is not intrinsically evil, and can help make scripts easier to read. Those of us who used to write batch files which did > nul use it as a reflex .. but used to supress an object which enumerates into 10s or 100s  of thousands of items, sending each of them down the pipeline _to do nothing with them_ makes no sense.

ili101:
It actually 30 times slower even on single item. But the point is that instead of thinking every time if it's harmful or not to use it it's easier just to get used to using something else

jhoneill:
I found my 500,000 cell sheet exported in 4 seconds instead of 8 so we're talking about microseconds which only matter when you have a big data set. You're right that one should get out of the habit of _automatically_ writing | out-Null , using $null = , or [void] is a _better default_, but if you look at something and say "that would be clearer if I put | out-null"  (or return $x ; or write-output $x) then that's the version to use _unless_ there is a big perf hit. But if they don't improve clarity, don't use them. 
I use all 3 interchangably, but deep down I don't like variables with special behaviours,  especially if they are undocumented (the about_automatic_variables help doesn't talk about assigning to $null  although the description from get-variable does) - that leaves me wondering if someone who is a expert in (say) AD or Skype for Business but has only basic PowerShell will understand what the script does.

ili101:
Yes I agree that it comes down to personal taste in the cases when performance is not impacted.
As you said above:
> Then we should just use one the others, not both :-)
If I have 3 options and I can use one of them for all cases I think it's the cleanest solution.

dfinke:
Nice. Good point on special variables too. I need to add that choice point for when I build things. Others will be reading my code. Tho I do like `echo a b c` to create an array. Excellent whipupitude factor :)

PR #590, opened 2019-05-01
Fixed case on file names, and applied a new rules from script analyzer

jhoneill:
Very little functional change - remove-worksheet supports -confirm, and I changed 3 or 4 instances of ::new() to new-object because that will work with PS 4  (we don't expressly want to support 4, but at the same time there's no sense in breaking 4 if we don't need to)
I've made some changes to either bring the code into  line with PSScriptAnalyzer 1.18 guidlines or to tell it where it doesn't need to worry. 
Also fixed a test which didn't work in the appveryor environment.

### DIFF
--- a/AddConditionalFormatting.ps1
+++ b/AddConditionalFormatting.ps1
@@ -185,7 +185,7 @@
             $Address = "$($Address.Row):$($Address.Row)"
     }
     elseif ($Address -is [OfficeOpenXml.ExcelColumn]) {
-        $Address = [OfficeOpenXml.ExcelAddress]::new(1,$address.ColumnMin,1,$address.ColumnMax).Address -replace '1',''
+        $Address = (New-Object 'OfficeOpenXml.ExcelAddress' @(1, $address.ColumnMin, 1, $address.ColumnMax).Address) -replace '1',''
         if ($Address -notmatch ':') {$Address = "$Address`:$Address"}
     }
     if ( $Address -is [string] -and $Address -match "!") {$Address = $Address -replace '^.*!',''}
--- a/AddDataValidation.ps1
+++ b/AddDataValidation.ps1
@@ -69,9 +69,9 @@
     }
     else {
         #We should accept, a worksheet and a name of a range or a cell address; a table; the address of a table; a named range; a row, a column or .Cells[ ]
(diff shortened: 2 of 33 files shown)


dfinke:
Did the new PSSA catch the casing?

jhoneill:
No, it didn't. I installled the  module in cloud-shell and 3 or 4 files were not found. All the changes outside the PSM1 came from PSSA. I've left a couple of the things it doesn't like for now.  Your `if ($x = do-something)` trick gets queried, and I didn't want to disable checks for that, and there is at least one place where there is a plural name which could be re-thought but I'm leaving alone for now.

dfinke:
# Might be a useful report to run on commits

```ps
$xlfile = "$env:TEMP\ScriptAnalyzer.xlsx"
Remove-Item $xlfile -ErrorAction SilentlyContinue

$p = @{
    AutoSize          = $true
    AutoFilter        = $true
    Show              = $true
    IncludePivotTable = $true
    IncludePivotChart = $true
    PivotRows         = 'Severity', 'RuleName'
    PivotData         = @{RuleName = 'Count' } 
    ChartType         = 'BarClustered '
    Activate          = $true
}

Invoke-ScriptAnalyzer . | Export-Excel $xlfile @p
```

image

jhoneill:
@dfinke Genius ! The highest praise I can offer it is to put it straight into my presentation for later ...

dfinke:
# Need to find a different word :-)

You don't do that! Please amplify. Probably will tweet this.

**Plagiarism** is the "wrongful appropriation" and "stealing and publication" of another author's "language, thoughts, ideas, or expressions" and the representation of them as one's own original work.

jhoneill:
@dfinke  
### Research 
(see Tom Lehrer)

jhoneill:
@dfinke I've checked in an example named analyze-this.ps1 
cd ..\GitHub\ImportExcel\
>"." , (dir 'C:\Program Files\WindowsPowerShell\Modules\ImportExcel\')  | .\examples\ScriptAnalyzer\Analyze_this.ps1

produces this which compares the quality over versions 
image

dfinke:
Thanks! Was going to ask. I'll check in the the other too.

dfinke:
Hmm. Don't see it.

Issue #596, opened 2019-05-16
Importing Multiple Tables from one Sheet is Slow


brovoldt:
I am working on trying to load a workbook into a database.  The workbook has multiple sheets and some of those sheets have many tables within each sheet.  I am able to pull the data the way I want however due to the size of the file of about 30MB it seems to be very slow each time I am pulling a new set of data.  Is there a more efficient way to pull multiple sheets or multiple tables per sheet to increase the speed?  

An example of the code I am using is below but this is only for 2 of the sheets in a workbook that has roughly 60 sheets.  Each table from each sheet is taking about 20-30 seconds and the number of rows is about 85 and the number of columns is about 10-12.  This example code takes about 5 minutes to complete.

$Premiums_Columns = @("IssueAge", "MNS", "MS", "MC", "FNS", "FS", "FC", "UNS", "US", "UC")
$Premiums_w_Pref_Columns = @("IssueAge", "MSPP", "MPP", "MNS", "MS", "MC", "FSPP", "FPP", "FNS", "FS", "FC", "UNS", "US", "UC")

$Sheets = @"
SheetName,PremType
100PL Prems,1
20PL Prems,2
"@ | ConvertFrom-Csv  

$dataArray = @()
$dataArray=$null 
#Premium Rates 
foreach ($Sheet in $Sheets) {
    #Bands
    $Prems_10K = Import-Excel -Path $SourceFile -WorksheetName $Sheet.SheetName -StartRow 7 -startcolumn 1 -endcolumn 10 -DataOnly -HeaderName $Premiums_Columns

    $Prems_25K = Import-Excel -Path $SourceFile -WorksheetName $Sheet.SheetName -StartRow 7 -startcolumn 11 -endcolumn 20 -DataOnly -HeaderName $Premiums_Columns

    $Prems_50K = Import-Excel -Path $SourceFile -WorksheetName $Sheet.SheetName -StartRow 7 -startcolumn 21 -endcolumn 30 -DataOnly -HeaderName $Premiums_Columns

    $Prems_100K = Import-Excel -Path $SourceFile -WorksheetName $Sheet.SheetName -StartRow 7 -startcolumn 31 -endcolumn 44 -DataOnly -HeaderName $Premiums_w_Pref_Columns

[...]

PR #602, opened 2019-05-24
Allow import Excel to take packages not just paths

jhoneill:
As it says in the title. I've added an ExcelPackage parameter and switched the open and close file logic when there is a path to use the Open-ExcelPackage and Close-ExcelPackage function. Also some small tweaks to help text.

### DIFF
--- a/ImportExcel.psm1
+++ b/ImportExcel.psm1
@@ -74,7 +74,11 @@ function Import-Excel {
 
    .PARAMETER Path
        Specifies the path to the Excel file.
-
+   .PARAMETER ExcelPackage
+       Instead of specifying a path provides an Excel Package object (from Open-ExcelPackage)
+       Using this avoids re-reading the whole file when importing mutliple parts of it.
+       To allow multiple read operations Import-Excel does NOT close the package, and you should use
+       Close-ExcelPackage -noSave to close it.
    .PARAMETER WorksheetName
        Specifies the name of the worksheet in the Excel workbook to import. By default, if no name is provided, the first worksheet will be imported.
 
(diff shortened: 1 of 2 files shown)


dfinke:
## Am I using it correctly

This is only a second faster. I guess the perf boost is on reading parts of each sheet of the already "opened" xlsx?

[testData.xlsx](https://github.com/dfinke/ImportExcel/files/3219422/testData.xlsx)

```powershell
    $xlfile = ".\testData.xlsx"
    foreach($sheet in "100PLPrems","20PLPrems") {
        $null = Import-Excel -Path $xlfile -WorksheetName $sheet
    }
}

function readAsPkg {

    $xlfile = ".\testData.xlsx"    
    $excel = Open-ExcelPackage $xlfile
    foreach($sheet in "100PLPrems","20PLPrems") {
        $null = Import-Excel -ExcelPackage $excel -WorksheetName $sheet        
    }
    Close-ExcelPackage $excel -NoSave
}

(Measure-Command { readEachTime }).TotalSeconds # 8.6
(Measure-Command { readAsPkg }).TotalSeconds    # 7.2
```

jhoneill:
That's pretty much expected. The smaller the fraction of the total which is read by import, and the more imports from the same file the bigger the delta. The open time in EPPlus for my "big" test workbook is about 3.5 seconds, and the import time after that is about 12 seconds for the biggest sheet (500K cells). 
So two passes I'd drop from ~32 to ~28 seconds.  But that's only a 5MB  file  

In #596 the talk was of a 30MB sheet so that would be ~20 seconds to open and each import is only ~1000 cells , so most of the time is going to be in the open and he talked of 60 sheets to process, so I could see this shaving ~20 minutes off the total time.  Other cases it might only save

PR #619, opened 2019-06-20
Organizing cmdlets

mhokanson:
I moved the cmdlets into their own folder to get them out of the root of the project and used a function I learned/borrowed from https://github.com/gscales/Exch-Rest to import anything in the cmdlet folder into the module. This should also make it easier to add new cmdlets to the project, as you now won't have to track whether or not the .ps1 is explicitly named in the .psm1 file.

I've tested locally with the Export-Excel cmdlet and things appear to work.


dfinke:
Thanks, not going to do this at this time.

Issue #627, opened 2019-07-04
Import-Excel relative path needs resolving


klemmt:
Was using the module in version 5.1.0 up to yesterday.
With the update to 6.2.1 relative paths stopped working for function Import-Excel.
Was introduced with a change to Version 6.

**Example:**
``PS c:\>Import-Excel -Path ".\someDocument.xlsx"``

**Error:**
```
New-Object : Exception calling ".ctor" with "4" argument(s): "Could not find a part of the path 'C:\Users\#USERNAME#\someDocument.xlsx'."
At C:\Program Files\WindowsPowerShell\Modules\ImportExcel\6.2.1\ImportExcel.psm1:349 char:23
+ ...   $stream = New-Object -TypeName System.IO.FileStream -ArgumentList $ ...
+        
```         
Reason System.IO.FileStream object needs a fully defined path.
Section of the code that was removed with from Version 5.1.0 on ``line 326``
```PS
$Path = (Resolve-Path $Path).ProviderPath
```

Please reintegrate into ImportExcel.psm1, before ``line 349``, for a fixe I have now integrated the line above into my code when calling the function.

dfinke:
https://github.com/dfinke/ImportExcel/releases/tag/v6.2.2

PR #628, opened 2019-07-04
Added resolve-path and tests. #627; Commented out write-warning

dfinke:


dfinke:
@jhoneill Would you be up to do a review?

jhoneill:
I'll take a look

jhoneill:
Just one thing. We can pipe filenames into import , and the check for valid path and extension looks like it only processes the first time, `if (!$firstTimeThru) ` so I would drop that. Otherwise, looks great to me.

PR #640, opened 2019-07-25
Count columns and line?

sam874:
Hello,

Reading the number of columns and rows in a WorkSheetName ?

Thank you

Best regards


dfinke:
@sam874 Could you post your question in the issues section and then I can try to get you an answer.

PR #659, opened 2019-08-16
Added Out-Excel Command

DomRRuggeri:
This is a function I set up because I found myself using a lot of the same parameters with Export-Excel, and just wanted a clean spreadsheet quickly.

Out-Excel expects input from the pipeline, and only requires a path. If .xls/.xlsx extension is not specified, .xlsx will be automatically added.

If no TableStyle is specified, Medium2 TableStyle is used.
If no WorksheetName is specified, "Data" is used.
Freeze Top Row, AutoSize, and AutoFilter are used by default.
WorksheetName is used for TableName by default.

Excel opens the file automatically after processing, unless the 'DontOpen' switch is used.
'Force' switch can be used to overwrite existing file path.

Syntax can be as simple as:
`Get-Service | Out-Excel C:\Temp\Services`

Or, as complex as:
```
Get-Service | Out-Excel -Path C:\Temp\Services-Processes.xlsx -Sheet "Services" -TableStyle Light3 -Force -DontOpen
Get-Process | Out-Excel -Path C:\Temp\Services-Processes.xlsx -Sheet "Proccesses" -TableStyle Light4
```

Thank you!
Dom Ruggeri


ili101:
I think this is redundant as we already have now mode that fulfills the same purpose:
`Get-Service | Export-Excel`
after this fixes https://github.com/dfinke/ImportExcel/issues/657 this is the same thing.

The only difference is that now mode use -AutoFilter by default and you propose using `-TableName Data` and -FreezeTopRow by default which is a personal preference.

In my personal preference -FreezeTopRow is unnecessary with -TableName as Excel already "Fixes" the column names in Table mode instead if showing A, B, C...
Regarding -TableName being use by default instead of -AutoFilter I support that but I don't know if we are the majority on this one?

dfinke:
I glanced at this early and it seemed like `Export-Excel -Now` and the `-Now` can be left off to get the same behavior. IIRC I built `-AutoFilter` first and that is why it's the default. Changing that to `-Tablename` feels better, it has a nicer layout. It's a breaking change but seems reasonable.

Allowing a `-WorksheetName` also seems reasonable. Specifying the `-Path` bypasses the `-Now` and  `-show:$false` does not work.

Seems like a bunch of work that could be troublesome and I'm not sure if any tests are in place to test this dance.

So, naming the file, the worksheet and getting a TableName. Feels like a bunch of effort to get right and test. Tablename with a random name seems straightforward, but the devil is in the details.

`Out-Excel` could be added as an example.

DomRRuggeri:
@ili101, @dfinke Thanks for reviewing this.
I see what you guys are saying, and I agree this can be accomplished with the -Now parameter, and other adjustments as @ili101 has proposed in #661 

Thanks!

dfinke:
@DomRRuggeri Thanks for the PR, much appreciated. Feel free to do more, plus, ask questions in the issues of the repo. This is good collaboration.

PR #661, opened 2019-08-17
Out-Excel and -Now

ili101:
**Export-Excel** Allows the use of -Path with -Now.
**Export-Excel** Allows override of -Path ,-Show, -AutoSize and -AutoFilter when using -Now.
**Out-Excel** Is a Proxy function for `Export-Excel -Now [-OtherParams]`.
``` PowerShell
# Basic (-Now).
Get-Service | Out-Excel
# Override all the default -Now params including Path.
Get-Service | Out-Excel -Path "$env:TMP\File.xlsx" -Show:$false -AutoSize:$false -AutoFilter:$false
# Switch -AutoFilter to -TableName.
Get-Service | Out-Excel -TableName 'Data'
```
This resolves Issue https://github.com/dfinke/ImportExcel/issues/657.
Regarding https://github.com/dfinke/ImportExcel/pull/659. How I set it now the default is -AutoFilter, you can switch to a table with `-TableName 'Data'` or use none with -AutoFilter:$false.
If I change the default from -AutoFilter to -TableName we will be able to switch to -AutoFilter but I don't know how to make it switch to none as I can't do `-TableNamer:$false` as -TableNamer is a string. Any suggestions?


ili101:
I changed the -Now default from -AutoFilter to -TableName
``` PowerShell
# Table (Auto name)
Get-Service | Out-Excel
Get-Service | Out-Excel -TableName ''
# Table (named)
Get-Service | Out-Excel -TableName 'Data'
# AutoFilter
Get-Service | Out-Excel -AutoFilter
# None
Get-Service | Out-Excel -TableName:$false
Get-Service | Out-Excel -TableName:$null
Get-Service | Out-Excel -TableName $false
Get-Service | Out-Excel -TableName $null
```
I needed to remove the type from the -TableName parameter for it to accept $false as [bool]

ili101:
@dfinke update me on what you think regarding the -TableName -AutoFilter swap and if you want me to move Out-Excel to the examples and add some instructions on how to tweak it.

dfinke:
Be great to move `Out-Excel` to the Examples folder, add a folder `CustomizeExportExcel`? 
The idea of defaulting the `TableName` that @DomRRuggeri had I liked.

Are there tests that exercise the parameters/parametersets?

ili101:
Added ParameterSets tests and moved Out-Excel to Examples\CustomizeExportExcel

dfinke:
@DomRRuggeri Thanks for the Out-Excel, good idea, conversation and thanks @ili101 for the refactor and especially the tests

@ili101  Are you on twitter? 

I'm going to update the readme, I'll ask you to double check, make sure I got everything.

ili101:
Not really posting on twitter.
Will be nice if also you and @DomRRuggeri give a look to the Out-Excel explanation I wrote to see if it and the hypothetical example I come out with are clear. and improve if needed.

dfinke:
I did take a quick look. I liked the `DynamicParam`, going to do a deeper dive when I can.

ili101:
The DynamicParam part is actually a modification of what @jhoneill done in https://github.com/dfinke/ImportExcel/blob/master/Send-SqlDataToExcel.ps1.
The difference is that it's simpler as it doesn't need the part that strip the parameter sets so it can take the parameters as is and add them.

dfinke:
Yup, cool stuff, now there's three places with `DynamicParams` https://github.com/dfinke/ImportExcel/blob/master/New-ConditionalFormattingIconSet.ps1#L39

PR #664, opened 2019-08-25
Fixes issues with newer EPPLus.

jhoneill:
Which were mostly round the tests, Also some bits on Protect worksheet.


jhoneill:
Some of the test  environments aren't setting $Temp so attempts to create fails go south ...

dfinke:
I'll disable the new pipeline for now. Maybe it's a `\` vs '/'.

ili101:
There is no $env:tmp on Linux
[IO.Path]::GetTempPath()

ili101:
Hi. I tried to integrate https://github.com/ili101/Module.Template and it went smoothly.
AppVeyor and Azure DevOps with badges per test, Artifacts and test results published.
image
You can optionally set up manual publish in Azure DevOps from an artifact.
Can F5 from VSCode to run tests.

The complicated part was the Linux tests:
First the casing of some of the files name needed to be fixed as Linux is case sensitive, There are 3 files lists: Install.ps1, ImportExcel.psm1, and filelist.txt that needed to be compared and fixed. (I noticed that there are 2 orphaned files Export-charts.ps1 and GetExcelTable.ps1 that are not referenced anywhere, not sure if intentional).
$env:TEMP needed to be replaced everywhere. in Export-StocksToExcel.ps1 I replaced it with `[IO.Path]::GetTempPath()` and in all the tests with [`TestDrive:`](https://github.com/pester/Pester/wiki/TestDrive). The hard part is that TestDrive: Is cleared when you go out of the test context it is set in, which is a good thing but because the why some of the tests was designed like creating a file in test 6 and using it again in test 9 some of the file creation code needed to move up a level to cover multiple tests without being cleared in the middle.

After this the Windows PowerShell 5.1, 6 and 7preview test are passing. The Linux test have only 5 fails and the macOS 6 fails.

ili101:
https://ci.appveyor.com/project/ili101/importexcel/builds/26968098/job/7321fellfp7pbc4u
Looks like the remaining errors on Linux are in the tests and not the module:
`Get-Service` and `Get-CimInstance` are missing in Linux.
`rm` alias need to be replace by `Remove-Item`.

The macOS one I don't know
```
  Describing Set-ExcelColumn, Set-ExcelRow and Set-ExcelRange
    [-] Error occurred in Describe block 0ms
      DllNotFoundException: Unable to load DLL 'libgdiplus': The specified module could not be found.
      TypeInitializationException: The type initializer for 'Gdip' threw an exception.
      MethodInvocationException: Exception calling "AutoFit" with "0" argument(s): "The type initializer for 'Gdip' threw an exception."
      at Set-ExcelRange<Process>, /Users/vsts/.local/share/powershell/Modules/ImportExcel/6.2.4/SetFormat.ps1: line 233
      at <ScriptBlock>, /Users/vsts/agent/2.155.1/work/1/s/__tests__/Set-Row_Set-Column-SetFormat.tests.ps1: line 146
```

jhoneill:
I think is good to merge now, we have some long standing issues with things in the test which don't run in mac and linux. I would have sworn I'd run the tests in wsl and azure shell. Maybe I just loaded the module. Do you want to give me a day or two to go through the tests on wsl ?

ili101:
I basically have the Linux tests part done. When you are finished with the EPPLus merge If you guys want I can make a PR.

jhoneill:
The only thing for the readme is "Set-WorksheetProtection" is now switched on. 
I might roll to 6.5 and keep V7 for when PS 7 , new .Net core, and possibly EPPlus updates. Or a total eclipse which ever comes first. 
I made a change to make set-Excel range more friendly when autosizing on non-windows platforms, and I've fixed the tests. 
Another PR will be hard on the heels of this one.

PR #691, opened 2019-10-07
ScriptAnalyzer

ili101:
Module.Template update to [2.0.4](https://github.com/ili101/Module.Template/blob/master/CHANGELOG.md)
* Moved Template CI files from `__tests__` folder to `CI` folder make it more organized.
* A ScriptAnalyzer report is generated and saved in Azure Pipelines > Artifacts > Source > `ScriptAnalyzer.xlsx`. In a PR the report will include the PR analysis and the target branch analysis so you can compare.

This PR is basically implementation of your example https://github.com/dfinke/ImportExcel/blob/master/Examples/InteractWithOtherModules/ScriptAnalyzer/Analyze_this.ps1
Also mentioned https://github.com/dfinke/ImportExcel/pull/590

Download:
image

Report:
image

Report PR mode:
image


dfinke:
Took a quick look, looks good, will take a deeper look. I'm thinking of dropping appveyor. 

@jhoneill Any thoughts?

ili101:
AppVeyor advantages is that if you need to  troubleshoot something you can SSH/RDP to the VM. And you can change the powershell core version (AppVeyor is used now to test on PS7 preview 4).
The disadvantage is that the free version is very slow.

jhoneill:
Same as you. Sounds good. I'll have more of a look and will either use it other places or will see something wrong :-)

dfinke:
@ili101 Good points. Haven't needed that. Wonder when/if Azure DevOps will have that? I prefer the performance Azure DevOps has. It's not really needed for the frequency of updates to this repo though.

PR #713, opened 2019-11-04
Fix import excel headers

dfinke:


dfinke:
@jhoneill, before  I go down the rabbit hole. Looks like checking for `HeadingName` and using `$Rows = $StartRow..$EndRow` solves it. 

Put some basic testing in place to check offsets.

Thoughts? Before I go deeper.

jhoneill:
I think this is by design - not that the design was _right_ 

Basically we have two choices. "There are no headers" and "Replace the headers with ...". 
If I understand what this change does, anything which has headers which need to be replaced will get them as a row of data.  Really if we are going to fix the "I have no headers , use these names as headers" we need to enable both the options together.

ili101:
If you already digging in this subject can you see if you can implement the features I suggested in https://github.com/dfinke/ImportExcel/pull/579

I was thinking maybe doing something like this to address this problem +  https://github.com/dfinke/ImportExcel/pull/579 + more option we will want to add now or later:
``` powershell
# This is an incomplete mockup
# SYNTAX: New-HeaderDefinition [-MyTableHasHeaders] [-IncludeUndefinedHeaders]
$Headers = New-HeaderDefinition -MyTableHasHeaders
# SYNTAX set OriginalIndex : Add-HeaderDefinitionColumn -OriginalIndex <int> -NewName <string> [-NewIndex <int>]
# SYNTAX set OriginalName  : Add-HeaderDefinitionColumn -OriginalName <string> -NewName <string> [-NewIndex <int>]
$Headers | Add-HeaderDefinitionColumn -OriginalIndex 1 -NewName 'New Name'
$Headers | Add-HeaderDefinitionColumn -OriginalName 'C3' -NewName 'Was C3' -NewIndex 1

Import-Excel -Headers $Headers

<#
Will replace
| C1 | C2 | C3 |
| A  | B  | C  |
with
| Was C3 | New Name |
| C      | A        |
#>
```
This way we can keep the number of parameters on Import-Excel to the minimum while giving more flexibility, but it requires more then one line of code to use.
What do you think? good or overkill?

dfinke:
I'm all for doing the simplest change to get the desired outcome of the initial bug report, with an eye towards no breaking changes.

Does the code change in the PR satisfy that?

jhoneill:
I think the simplest way is to  change this bit at line 345 in importexcel.psm1
```
                if ($NoHeader) {
                    $i = 0
                    foreach ($C in $Columns) {
                        $i++
                        $C | Select-Object @{N = 'Column'; E = { $_ } }, @{N = 'Value'; E = { 'P' + $i } }
                    }
                }
                elseif ($HeaderName) {
                    $i = 0
                    foreach ($H in $HeaderName) {
                        $H | Select-Object @{N = 'Column'; E = { $Columns[$i] } }, @{N = 'Value'; E = { $H } }
                        $i++
                    }
                }

```
Put if($headerName) first and then elseif ($Noheader). That will set the names then it can either be run with startrow 2 (or top of data +1 ) or with -NoHeader (which means simplifying the parameter sets. Update help & tests, and job done. 

I

dfinke:
Doing it KISS. @jhoneill Looks like the `-DataOnly` still works on skipping empty rows, but not empty column. I'll check it out, once done, I'll merge.

Any thoughts on the tests etc.? I didn't touch the parameter sets. I think the help looks good.

Issue #718, opened 2019-11-08
Import-Excel accepts the PipelineVariable param, but the variable stays empty.


lucdekens:
Looks like this PipelineVariable on the Import-Excel cmdlet might not be working as expected.
The variable seems to stay empty.

Tested this on 6.5.2
![2019-11-08_14-32-23](https://user-images.githubusercontent.com/7760649/68481081-8d47bd80-0236-11ea-81ac-6f84504159b5.jpg)

Boe Prox did a post on how to implement the PipelineVariable parameter.
See https://mcpmag.com/articles/2016/12/15/pipelinevariable-common-parameter.aspx

ili101:
I see that someone opened an issue on it 17 days ago https://github.com/PowerShell/PowerShell/issues/10932
Basically if you want -PipelineVariable to work you can't use Begin, Process, End as it will work only on the first thing in the pipeline.
👍 on https://github.com/PowerShell/PowerShell/issues/10932 and maybe they will tell us if it's a bug or intended behavior.

lucdekens:
I'm not too sure if this is applicable in this case.
The sample I used above is not using an advanced function, nor any Begin/Process/End blocks.

ili101:
Import-Excel itself is an advanced function with Begin/Process/End.
For this to work (if not fixed in newer version of PowerShell at some point) we need to rewrite Import-Excel without Begin/Process/End

ili101:
Actually as the only Pipeline parameter is -Path I just removed the Begin/Process/End and looks like it works without any side effects that I can see, try loading this https://gist.github.com/ili101/a8d7b3bfedad159cfc6f667d88b6395c Import-Excel and try your command again, it should work.

lucdekens:
Confirmed, that does the trick

dfinke:
Thanks @ili101 do the unit tests work with that change? Can you add some to handle this scenario and do a PR?

PR #726, opened 2019-11-17
support PipelineVariable

ili101:
This is for https://github.com/dfinke/ImportExcel/issues/718
In order to workaround probable PowerShell bug https://github.com/PowerShell/PowerShell/issues/10932 I removed `Begin` and `Process` from `Import-Excel` and added a test for it.
(Ignore the big GitHub "File changes", it just detecting all the indentation changes badly, view changes in VSCode instead)


dfinke:
Hitting up against PS bugs, nice. Do you know, is this fixed in core etc?

ili101:
I'm on the last preview and it have the same problem

dfinke:
@ili101 thinking of merging @jhoneill changes, then these.

jhoneill:
If we don't have a process block what happens to scenarios like 
`dir timesheets\*.xlsx | import-excel  ...`

dfinke:
Yes. Need to add tests for that scenario first.

ili101:
Sorry, didn't know it was a valid scenario, probably not going to work.
I can add a test and fix it in a few hours when I get home.
Thank you @jhoneill nice catch

ili101:
Did some testing on the existing code, The scenario of multiple paths is not really supported right now
```powershell
Remove-Item -Path $Path1 -ErrorAction SilentlyContinue
Remove-Item -Path $Path2 -ErrorAction SilentlyContinue

$Path1 = Join-Path ([IO.Path]::GetTempPath()) 'test1.xlsx'
$Path2 = Join-Path ([IO.Path]::GetTempPath()) 'test2.xlsx'

# 1 column 2 rows
@(
    [PSCustomObject]@{F1I1 = 'a'}
    [PSCustomObject]@{F1I1 = 'b'}
) | Export-Excel -Path $Path1
# 2 columns 1 row
[PSCustomObject]@{
    F2I1 = 'c'
    F2I2 = 'd'
} | Export-Excel -Path $Path2

$data = $Path1, $Path2 | Import-Excel

'Table Format:'
$data
'List Format:'
$data | Format-List
```
Output:
```
Table Format:

F1I1
----
a
b


List Format:


F1I1 : a

F1I1 : b

F2I1 : c

F2I1 :
```
1. all of the lines from the 2 files are merged into 1 array.
2. The number of columns is set by the first file, if the second file have more columns (In this example)  they are discarded.
3. The number of rows is set by the first file, if the second file less more rows (In this example) they are added as empty.

So we need to decide if we want to support this scenario and if so we need to decide what it will do like if to apply all the input parameters to all of the files and how to output the results? as one big array or nested array for every file.

Also table format (Print to console) of the output will not be usable as PowerShell will distort it, but it's not something we can control:
https://stackoverflow.com/questions/31441054/formatting-issues-missing-data-when-multiple-commands-use-select-object-within

dfinke:
I believe this works as I remember. Let's start here with tests to prove it out and expand it when more scenarios are requested. I like the idea of pipeline variables, to be honest, never used them.

dir C:\Temp\testdata?.xlsx |Import-Excel

[testData1.xlsx](https://github.com/dfinke/ImportExcel/files/3877518/testData1.xlsx)
[testdata2.xlsx](https://github.com/dfinke/ImportExcel/files/3877519/testdata2.xlsx)


```
cola colb
---- ----
   1    3
   2    4
   5    7
   6    8
```

ili101:
Yes merging multiple file with exactly the same columns and exactly the same number of lines works but what are the chances of someone needing something like that? This is more confusing then helpful.

dfinke:
As in James' example, `dir timesheets\*.xlsx`, I'd expect all the timesheets from all employees and they'd be the same in terms of columns. Works for me.

ili101:
Added the 3 tests. same columns & rows count from pipeline and param and PipelineVariable.
Before multiple paths was supported by pipeline, but not from `-Path`. and `-PipelineVariable` didn't worked.
Now all 3 works.

dfinke:
@ili101 What was the workaround you needed to do initially that now works?

ili101:
Initially I just removed the Begin/Process blocks. that make the function behave like everything inside it is in an "end" block so if the pipeline is used you only get the last item in the pipeline.
That's ok if you only take one item from the pipeline, but as we want to support multiple paths that's not good enough.
So I use [$input](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_automatic_variables?view=powershell-6#input) that basically collect all the pipeline input that is passed to the function, then use it if exists:
```powershell
    if ($input) {
        $Paths = $input
    }
    elseif ($Path) {
        $Paths = $Path
    }
    else {
        $Paths = ''
    }
```
 and then replace where the "Process" blocks was with a `foreach ($Path in $Paths)` loop.

Issue #720, opened 2019-11-15
Incorrect table when supplying DataTable with zero rows


techabilla:
Sending a DataTable with columns, but no rows, should result in an empty table with just the header row, but what I'm seeing is a table with no header row, and one data row that contains the column names.

Expecting to see:
![image](https://user-images.githubusercontent.com/32886354/68911776-f9ca3b80-07a5-11ea-852b-bb48799d767f.png)

Seeing:
![image](https://user-images.githubusercontent.com/32886354/68911786-02227680-07a6-11ea-84a3-a77108cf05c4.png)

My code:
```
            $exportParams = @{
                Path = $outputPath
                InputObject = $dataSet.DataTable
                WorksheetName = $dataSet.ExcelExport.WorksheetName
                TableName = $dataSet.ExcelExport.TableName
                HideSheet = $dataSet.ExcelExport.WorksheetName
                ClearSheet = $true
                ExcludeProperty = @('RowError', 'RowState', 'Table', 'ItemArray', 'HasErrors')
            }
    
            $xlPkg = Export-Excel @exportParams -PassThru
```

Having looked at the Export-Excel code I'd guess this is an EPPlus issue, but thought I'd post in case there is a simple workaround.

Great module btw; thanks for making my life easier!

techabilla:
Code to demonstrate the problem

```
$d = new-object system.data.datatable
$d.Columns.Add("column1")
$d | export-excel
```
... as opposed to...
```
$d = new-object system.data.datatable
$d.Columns.Add("column1")
$d.Rows.Add("value1")
$d | export-excel
```

techabilla:
Have struggled to find a workaround for this. Adding a dummy row allows the table to be created, but I haven't been able to reliable 'empty' the table from PowerShell. At present I'm running a macro in the spreadsheet that checks for tables with dummy rows, and deletes the data range.

ili101:
Excel can't have a table without a row, So if you create one from MS Excel it will set up an empty row as @techabilla showed in "Expecting to see".

So we can add one row as shown in "Fix option 2" below which seems to work.
Or we can do it directly in LoadFromDataTable "Fix option 1" below (that's what I was using back when I head my module fork before we added the DataTable support here) .

```powershell
# Setup
$DataTable = [Data.DataTable]::new('Test')
$null = $DataTable.Columns.Add('ID', [Int32])
$null = $DataTable.Columns.Add('Name')
[...]

PR #729, opened 2019-11-20
Fix for #720,  plus some linting

jhoneill:
Fix for #720 

Also found a lot of the examples tried to load the module with an empty catch block which was a big source of messages from script analyzer so addressed that.


Export-Excel.ps1
```diff
@@ -620,27 +623,36 @@
...
-                Write-Warning "Table name $($InputObject.TableName) is not unique, adding '_' to it "
-                $InputObject.TableName += "_"
-            }
-            #Insert as a table, if Tablestyle didn't arrive as a default, or $TableName non-null - even if empty
-            if ($null -ne $TableName -or $PSBoundParameters.ContainsKey("TableStyle")) {
-                $null = $ws.Cells[$row,$StartColumn].LoadFromDataTable($InputObject, (-not $noHeader),$TableStyle )
-                # Workaround for EPPlus not marking the empty row on an empty table as dummy row.
-                if ($InputObject.Rows.Count -eq 0) {
-                    ($ws.Tables | Select-Object -Last 1).TableXml.table.SetAttribute('insertRow', 1)
+            if ($Append) {
```
ili101: Maybe do `if ($Append -and $ws.Dimension)` like in line 553 (Or something better if you think of it)
So when you do something like in https://github.com/dfinke/ImportExcel/issues/730#issue-525769283 example:
```powershell
for ($i = 0; $i -lt 3; $i++) {
    Export-Excel -InputObject $DataTable -Path $Path -TableName 'Test2' -Append
}
```
The first -Append will be ignored

Export-Excel.ps1
```diff
@@ -620,27 +623,36 @@
...
+                if ($TableName) {
+                    Add-ExcelTable -Range $ws.Cells[$ws.Dimension] -TableName $TableName -TableStyle $TableStyle
                 }
             }
-            else {
-                $null = $ws.Cells[$row,$StartColumn].LoadFromDataTable($InputObject, (-not $noHeader) )
+            else  {
+                #Change TableName if $TableName is non-empty; don't leave caller with a renamed table!
+                $orginalTableName = $InputObject.TableName
+                if ($PSBoundParameters.ContainsKey("TableName")) {
```
ili101: I see you changed here the `if ($TableName)` to `if ($PSBoundParameters.ContainsKey("TableName"))`
This makes this example create auto named table "Table1" instead of a table named "MyName":
``` powershell
$DataTable = [Data.DataTable]::new('MyName')
$null = $DataTable.Columns.Add('ID', [Int32])
$null = $DataTable.Columns.Add('Name')
Export-Excel -InputObject $DataTable -Path $Path -TableName ''
```
Was this intentional?
When using `-TableName ''` I think using the name of the table and only if it's also empty auto generate name is preferable.
What do you think is better?

jhoneill: Yes. It was intentional, because $tableName is set elsewhere in the function we only want to over-ride the name on the datatable object with something which came in as a parameter if it is passed as a parameter. 

You can't have blank table name, and really the name should be validated as not null or empty.  I'll change that.
Most code samples I've seen over the years just use 
[Data.DataTable]::new() 
and your example would be better that way with  -tableName "MyName" as part of export excel.  (Unless the datatable was later going to be used by something which needed the name.) 
IF people care about the table name in excel, AND they don't control the name in a returned table AND they two names must match then -tableName $dataTable.Tablename works. So I tend view making the names match as a corner case anyway :-)

ili101: $DataTable.TableName can't be null it's always a string.
`if ($PSBoundParameters.ContainsKey("TableName"))` over-ride the datatable object name in more cases then `if ($TableName)` as `if ($TableName)` exclude empty string and ContainsKey includes it.
But as you say it's a far edge case so you can leave it like it is doesn't really matter

jhoneill: :-) 
That's correct. 
In _now_ mode, table is assumed. If -TableName is 

- **$false** (which can be written -Table:$false like a switch)  table is disabled
- **''** - Excel table is named table1,2,3 etc. 
- **Not specified** - use the DataTable's name if there is one, and table1,2,3 if there is not

In _Other_ mode  table is only created if name or style is specified. 

- **''** - Excel table is named table1,2,3 etc. 
- **Not specified**  (i.e. Table style was passed but Table name was not)- use the DataTable's name if there is one, and table1,2,3 if there is not

Issue #525, opened 2019-01-16
ConvertFrom-ExcelToSqlInsert does not escape single quotes


DerValle:
When I use the CmdLet on an Excel Sheet that contains text with single quotes, I get lot's of faulty quieries.
I would expect that the CmdLet makes sure that queries work regardless of the data quality.

Best regards and happy hacking

dfinke:
Please, could you post and example xlsx file

beckerben:
duplicate of #512

Issue #512, opened 2018-12-19
ConvertFromExcelToSQLInsert dealing with single quotes


jotaene:
Hi
Congratulations for your excellente job.
I found that ConvertFromExcelToSQLInsert.ps1 script doesn't deal well with words that contain a single quote like «I'm» because it should become «I''m» for SQL purposes.

In my local machine I turned ConvertFromExcelToSQLInsert.ps1 line 111 from ["'" + $record.$propertyName + "'"] into ["'" + (($record.$propertyName.ToString()) -replace "'","''") + "'"] and it seems to work.

Best regards

edwardmjackson:
This commit seems to have been lost in the master branch

PR #731, opened 2019-11-20
fix for #512 & #525

beckerben:
There is an issue with the ConvertFromExcelToSQLInsert where a quote in the text will not be converted properly and will cause invalid insert statements to be created, this commit makes the correction to handle the scenario.


beckerben:
will fix failed check

PR #732, opened 2019-11-20
Fix for #512 & #525

beckerben:


dfinke:
LGTM, @jhoneill?

jhoneill:
> LGTM, @jhoneill?

Yes, as far as a I can see "O'Neill" would break before and will be OK now. I don't imagine any side effects.

sthoen:
This bug is still present.  The 7.1.1 code is missing the #732 fix.

"'" + (($record.$propertyName) -replace "'","''") + "'"

dfinke:
@sthoen please open a new issue. This may take a bit. Need to upgrade to the latest version of Pester.

sthoen:
Thanks Doug, I have submitted the issue as #920.

Sebastiaan

PR #738, opened 2019-11-25
Refactor

jhoneill:


dfinke:
@jhoneill what were your thoughts on the conflicts?

jhoneill:
It looks like the CI script doesn't work with the files moved around. That's fixable (and and useful learning too).

ili101:
@jhoneill yes this is by design.
The concept behind this is to do the following steps:
1. CI Install the module and load it, this include only the files that are going to be published to the gallery.
2. CI Run the tests on the installed module.
3. You make the release from the installed module (@dfinke don't use this but it's the concept).

This has the following advantage:
You run the tests on the installed module in the exact same configuration as it will be published, so if there any unexpected faille like something with the path or the importing, or there are some files that you forget to include in the release the test will fail.
So the version you test on and the version you release to the gallery are identical which prevent release process related problems.

jhoneill:
@ili101  
Yes: you've moved things forward from just running tests to doing a proper CI build, that's a good thing and worth a small amount of extra work when the code is moved round on the scale I have done here. Now that the help is its own MD files they should probably be compiled in this process as well.

ili101:
@jhoneill Wow that is a big refactor.
Go to the file `Install.ps1` and change `$IncludeFiles` to:
```powershell
$IncludeFiles = @(
    '*.dll',
    '*.psd1',
    '*.psm1',
    'Charting',
    'en-US',
    'ExportedCommands',
    'Charting',
    'InferData',
    'Pivot',
    'Plot.ps1',
    'mdHelp',
    'ArgumentCompletion.ps1'
)
```
(If I missed something add it)
This will fix the tests (you can also execute `Install.ps1` locally and it will install the module for you locally).

Also just throwing in an idea, since we have so may files, I see it's commune to create a "Private" and  "Public" folders add put all the functions files inside, maybe it will be more organized that way?
https://github.com/RamblingCookieMonster/PSSlack/tree/master/PSSlack
https://github.com/PoshCode/ModuleBuilder/tree/master/Source

jhoneill:
@ili101 
Very few code changes but
1. All the help is in MD files for platyps and not as comments. That makes the files easier to work with but also lets someone translate the help
2. I'm trying to move to 1 file : 1 function , at least for the the working source files. I know some people like to have everything in the PSM1 file but it's easy to merge files automatically.  That's mostly done, a few need some work 
3. I've tried to fix as many of script analyzer's complains as I can. 
4. I fixed as many bit of inconsistent casing as I could. I probably introduced a lot of them in the first place :-) 
5. Public / Private directories, yes, that's the direction I'm going in. At the moment the name is ExportedFunctions, if there is any sort of standard like "Public" I'm happy to use it. 
The aims is to have a directory with
.psd /.psm, file, readme, licence and EPPlus.dll in the root then folders for 
en-us <and other help>
Exported / public functions
Helper  / private functions
Examples / images / testamonials

CI , Tests, and MD help files wouldn't be packed with the installable module.

dfinke:
Thanks @jhoneill, 7.0? Do you want to do the Public/Private directories for functions? I have no preference. For any project, when in Rome do as the Romans do.

All looks good on my machine. I have other modules that depend on ImportExcel. I installed this new version the tests run on my other ones with no issues (not that they fully exercise all the things).

dfinke:
@ili101 Do you want to do PR and update the message about `sudo apt-get` to do the update? I only got that issue trying a `Dockerfile` and GitHub Actions.

dfinke:
Rather than pipe `get-childitem ... | foreach`, is it worth doing a `foreach($i in $j)` in case this grows into lots of files?

ili101:
Is this considered done?
As I see some of the files has not been moved yet to the new folders.

dfinke:
Yes. More to come.

jhoneill:
These are the things I am aware of. 
```
                            Not Verb-Noun   Not In public   Help Missing
BarChart                    X               X               X
ColumnChart                 X               X               X
ConvertFrom-ExcelData                                       X
ConvertFrom-ExcelSheet                                      X
ConvertTo-ExcelXlsx                                         X
DoChart                     X               X               X
Export-ExcelSheet                                           X
Export-MultipleExcelSheets                                  X
Get-ExcelColumnName                                         X
Get-HtmlTable                                               X
Get-Range                                                   X
Get-XYRange                                                 X
Import-Html                                                 X
Import-UPS                                                  X
Import-USPS                                                 X
Invoke-AllTests                             X                 
Invoke-Sum                                                  X
LineChart                   X               X               X
New-ExcelStyle                                              X
PieChart                    X               X               X
Pivot                       X               X               X
Set-CellStyle                                               X
Set-WorksheetProtection                                     X
Test-Boolean                                X               X
Test-Date                                   X               X
Test-Integer                                X               X
Test-Number                                 X               X
Test-String                                 X               X


```
New-Plot is not in public as it can't be loaded on Downlevel versions 

EPPlus.dll could move to private or "assemblies" and out of root.
Azure-pipelines.yml can move to ci (but the build pipeline needs to be updated)
Export-charts.ps1, Export-StocksToExcel.ps1, and GetExcelTable.ps1: are they  examples, or should they be exported ?

Plus we (cough @dfinke ) should have a blitz on the open issues, and close all those which are fixed, answered questions or have just petered out and then focus on those that need fixing. 

I haven't checked the behaviour of help if the language isn't English and we could move some of the other text into a string resources file.

dfinke:
Yup. 

I experimented with the `*Chart` functions I think I saw in either numpy or scipy. It uses the `Test-*` functions to find out the data types of the data set passed and try to choose the X & Y to graph.

`New-Plot` also came from the Python libraries.

Issue #744, opened 2019-12-06
ConvertFrom-ExcelToSQLInsert StartRow raises error 


edwardmjackson:
Using ConvertFrom-ExcelToSQLInsert with StartRow raises the following error:
"ConvertFrom-ExcelData : A parameter cannot be found that matches parameter name 'StartRow'."

This is because ConvertFrom-ExcelData only supports the HeaderRow 

Suggested fix:
ConvertFrom-ExcelData HeaderRow parameter should have the following Aliases:

[Alias('StartRow', 'TopRow')]

edwardmjackson:
PR https://github.com/dfinke/ImportExcel/pull/745

PR #745, opened 2019-12-06
Fixes Issue #744

edwardmjackson:
Fixes call from ConvertFrom-ExcelToSQLInsert with StartRow parameter

### DIFF
--- a/Public/ConvertFrom-ExcelData.ps1
+++ b/Public/ConvertFrom-ExcelData.ps1
@@ -8,6 +8,7 @@ function ConvertFrom-ExcelData {
         [ScriptBlock]$scriptBlock,
         [Alias("Sheet")]
         $WorkSheetname = 1,
+        [Alias('StartRow', 'TopRow')]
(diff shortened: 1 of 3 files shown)


dfinke:
Had the thought, to add a test that proved this didn't work and then add the alias and the test should pass.

edwardmjackson:
I have added tests, and also remerged the code for https://github.com/dfinke/ImportExcel/pull/732 as it was blocking me

dfinke:
LGTM. @jhoneill ?

jhoneill:
It looks like it is backed out the move from comment based to md help.

beckerben:
I ran into the StartRow error and this fix does work, however i noticed what may be an unrelated issue that also specifying "Header" parm generates "Parameter cannot be processed because the parameter name 'Header' is ambiguous" so I renamed this parm to match "HeaderName" in both ConvertFrom-ExcelData and also ConvertFrom-ExcelToSQLInsert.

Issue #741, opened 2019-11-30
ConvertFrom-ExcelSheet enhancement


scriptingstudio:
### Issues:
- impossible to customize date format
- impossible to prevent formatting at all

### New parameters:
[string]$dateformat
[switch]$asText

### Changes in code:
```
$importparam = @{asText = $asText}
if ($dateformat) {$importparam['dateformat'] = $dateformat} 
foreach ($sheet in $targetSheets) {
    Write-Verbose "Exporting sheet: $($sheet.Name)"

    $params.Path = "$OutputPath\$($Sheet.Name)$Extension"

    Import-Excel $Path -Sheet $($sheet.Name) @importparam | Export-Csv @params
}
```

I would have added alias for the function, `[alias('ExportFrom-Excel')]`

jhoneill:
Slightly confused. 
There is already an as text option. Since people are bound to want some columns as text and  some not, it is used as -asText * for all columns or -astext col1,col2,colx* for specific columns. 

dates come back as dates. You format them the same as the date returned by any other command.

scriptingstudio:
'Astext' there is in import-excel not in ConvertFrom-ExcelSheet. Import-excel returns dates in default format, but I need it in custom format. That is why I have to customize ConvertFrom-ExcelSheet after module update.

jhoneill:
Sorry, I saw the import-excel in the body and didn't read the title which said you were talking about ConvertFrom-Excel sheet. Convert probably should take -AsText and pass it through to Import. 

Import-excel does not turn dates into strings. 
```
>import-excel C:\temp\test.xlsx

Date                Name
----                ----
06/01/2001 00:00:00 Lucy
21/12/2005 00:00:00 Peter
```
Above, PowerShell is applying default formatting for dates (UK  in my case) , below it is applying a custom format 
```
>import-excel C:\temp\test.xlsx | ft @{f="MMMM-dd-yy";e='Date'},Name

Date           Name
----           ----
January-06-01  Lucy
December-21-05 Peter
```

I'll give some thought to what is the best thing to do for the export. Possibly supporting a `-Properties`  parameter in the style of a` Select-Object` ; I know you want something which transforms dates, but someone else will want different things on different columns, and another person will want to format numbers. If I can make one change to solve the problem for all those cases that's better.

scriptingstudio:
Oh, I forgot that I have used a custom version of import-excel introduced -dataformat parameter because I needed 'dd-MM-yyyy' format.

Issue #737, opened 2019-11-22
ImportExcel, New-ConditionalText, color the full line


lea13000:
Bonjour,

I woul like, with Export-Excel, color the full line according to the value of a single column.
![image](https://user-images.githubusercontent.com/58072857/69429327-90ad6f80-0d33-11ea-925d-3b0c4d6f7e7a.png)
When B= Stopped, the full line is red.

For this, I would like to used this '=$B1="Stopped"'.

Here is the code : 

```
$ExcelParams = @{
    Path = $env:TEMP + '\Exceltest.xlsx'
    Show = $true
    Verbose = $true
}
Remove-Item -Path $ExcelParams.Path -Force -EA Ignore
$ct2 = New-ConditionalText -Range "A1:H8" -ConditionalType Equal  -ConditionValue  '=$B1="Stopped"' -ConditionalTextColor Red -BackgroundColor White
Get-Service | 
Select Name, Status, DisplayName, ServiceName |
Export-Excel @ExcelParams -ConditionalText $ct2 

```
But when I open Excel, it's looking for all the cells with **_=$B1="Stopped_** (Black) instead of  formula (Grenn) 
![image](https://user-images.githubusercontent.com/58072857/69428922-aa9a8280-0d32-11ea-853d-620e81b49420.png)

Thank you !

jhoneill:
Bonjour ! 

It need the type to be "Expression" otherwise it searches for "=$B1" as text

The formula is entered without the first = sign - I can't remember if we remove it so  

`$ct2 = New-ConditionalText -Range "A1:H8" -ConditionalType Expression  -ConditionValue  '$B1="Stopped"' -ConditionalTextColor Red -BackgroundColor White 
`
Should do what you need.

lea13000:
Thank you for your reply, 

Unfortunately... 
![image](https://user-images.githubusercontent.com/58072857/70418208-47fbf300-1a63-11ea-97d7-047578917022.png)

The argument "Expression" is not recognized.

jhoneill:
Hmmm.  It looks like we didn't add expression to the types allowed in New-Conditional text 


`$ct2 = @{range= "A1:H8"; RuleType='Expression' ; ConditionValue=  '=$B1="Stopped"'; ForegroundColor="Red" ; BackgroundColor="White"} `
Should work until we can fix that .  
The names are the same as the parameters for ADD-ConditionalFormatting.

lea13000:
[...]

Issue #754, opened 2019-12-13
Import-Excel // Dates with formatting are imported as numbers


AlexBillon07:
Powershell 5.1
ImportExcel 6.5.2

Hello,

Dates are imported as numbers when they are formatted in Excel Files.
I am not sure if this is a bug or not.

I have read the issue https://github.com/dfinke/ImportExcel/issues/289 and learned about the `DateTime.FromOADate(Double)` method. So I have a workaround but it would be nice to import formatted dates as dates with the Import-Excel cmdlet.

[Test.xlsx](https://github.com/dfinke/ImportExcel/files/3960584/Test.xlsx)

```
PS C:\Temp> Import-Excel .\Test.xlsx -WorksheetName 'WithoutFormatting'

Number Text Dates
------ ---- -----
     1 azae 13/12/2019 00:00:00
     2 fdz  14/12/2019 00:00:00
     3 1    15/12/2019 00:00:00
     4 azd  16/12/2019 00:00:00
     5      17/12/2019 00:00:00
     6
     7


PS C:\Temp> Import-Excel .\Test.xlsx -WorksheetName 'Formatted'

Number Text Dates
------ ---- -----
     1 azae 43812
     2 fdz  43813
     3 1    43814
     4 azd  43815
     5      43816
     6
     7

```

Best Regards,

Alex

jhoneill:
It is a known problem. 
Excel doesn't have dates. It just has numbers as days since 1 Jan 1900 converted to date and time. 
The EPPlus library recognizes dates if the cell uses one of the built in date formats (each format used in the sheet has an ID. Low numbers are assigned to built in formats and user formats get a higher number. Dates are 14,15,16,17). EPPlus doesn't look to see if the format is a combination of letters, numbers and separators which would make a date format, and you can imagine how slow that would be if we tried to do it for each cell. 
There are two options  one is to use FromOADate() 
The other is to import with   -AsText "Dates" (or whatever your column name is) this will preserve the text formatting which may be what you want; but if you want proper dateTime object from a  custom formatted column the FromOADate method is the best way.

AlexBillon07:
Yeah, I thought so. But I wanted to be sure.
Thanks.

PR #755, opened 2019-12-15
Latest updates

jhoneill:
Address #737 bug not allowing expression as a rule type
and #741 improvements to convert-fromExcelSheet - added MD help for this command at the same time
and #746 to check module manifest. (Further tidy up of .psd1 file)

Additional tests


jhoneill:
@dfinke  - I think this is ready when you are. 
I have one other possible thing in the queue which is to look at "-AsDate" for importExcel (see e.g. #754 ) but it needs a bit of time devoted to it which I don't expect to have in the next few days.

PR #782, opened 2020-02-14
Remove Write-Debug Performance Outputs

JustinGrote:
These outputs are unnecessary and can confuse someone observing other debug output.


dfinke:
@jhoneill Is it worth adding a param switch to output this info in addition to the `write-debug`?

jhoneill:
> @jhoneill Is it worth adding a param switch to output this info in addition to the `write-debug`?

It's a judgement call really. I've never felt the need for the performance and timing stuff myself (on that much I agree with Justin that it isn't _needed_, but nothing in debug is needed... I don't think it is confusing)  but there was a time (I think) when you found it useful.  
My instinct is not to have multiple debugging modes. -Verbose originally output something for every cell, and I got a big perf boost out of not calling write-verbose when there wasn't going to be any output. This info might sit better in verbose than in debug, but I'd say one of those or not at all.

PR #793, opened 2020-02-29
Spike to fix #792

### LINKED ISSUE #792: Odd behaviour on the return of Import-Excel function
Hello,

When we import an Excel file containing only 1 row of headers or 1 row of headers and 1 row of data, the function Import-Excel behaves in an odd manner,  making the task of controling the size of the imported array of data a bit difficult.

With the attached file test.xlsx, you can test yourself the 2 cases for which the return of the function is a bit odd:

**testing scenario 1**:  sheet 'no data'  

`$rows = Import-Excel -Path .\test.xlsx -WorksheetName 'no data'`

In this case, the return of the function is an Object[] with 2 rows!  The 1st is empty and the 2nd contains the headers.

**testing scenario 2**:  sheet '1 row'

`$rows = Import-Excel -Path .\test.xlsx -WorksheetName '1 row'`
In this case, the return of the function is PSCustomObject.

For sheets with 2 or more rows, we get again returned objects of type Object[], which is the expected behaviour.

My questions/requests:

1. Couldn't we have the function returning always the type Object[] even for scenario 2?

2. In scenario 1, couldn't we have the function returning simply an empty object[] for which we simply would test the property count in order to check the empty contents?

[test.xlsx](https://github.com/dfinke/ImportExcel/files/4267910/test.xlsx)

dfinke:
@jhoneill this "works". Might be opening a can of worms. Added a couple tests. One for the issue and the other using the `-NoHeader` switch. 

If there is no data, returns null, as does `ConvertFrom-Csv`

More tests? Rework? Thoughts?

### DIFF
--- a/Public/Import-Excel.ps1
+++ b/Public/Import-Excel.ps1
@@ -1,8 +1,8 @@
 ﻿function Import-Excel {
-  [CmdLetBinding()]
-  [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingPlainTextForPassword", "")]
-  [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSPossibleIncorrectUsageOfAssignmentOperator', '', Justification = 'Intentional')]
-  param (
+    [CmdLetBinding()]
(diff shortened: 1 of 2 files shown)


jhoneill:
Looks OK   - it's specific enough that I don't think it's going to let any worms out !

PR #835, opened 2020-05-01
fixes #832

### LINKED ISSUE #832: Importing module on macOS results in failure due to incorrect placement of `finally` block
On the PSM1 for ImportExcel v7.1.0, the last `finally` statement is wrapped within the `catch` statement preceding it instead of just after, resulting in the module failing to import if the `if` statement checking for `($IsLinux -or $IsMacOS) -or $env:NoAutoSize` returns `$true`:

image

***

Should be a simple fix, PR coming in shortly for it.

scrthq:
fixes #832

### DIFF
--- a/CI/CI.ps1
+++ b/CI/CI.ps1
@@ -13,7 +13,10 @@ param
     # AppVeyor and Azure - Upload module as AppVeyor Artifact.
     [Switch]$Artifact,
     # Azure - Runs PsScriptAnalyzer against one or more folders and pivots the results to form a report.
-    [Switch]$Analyzer
+    [Switch]$Analyzer,
+    # Installs the module and invokes only the ModuleImport test.
+    # Used for validating that the module imports still when external dependencies are missing, e.g. mono-libgdiplus on macOS.
+    [Switch]$TestImportOnly
 )
 $ErrorActionPreference = 'Stop'
 if ($Initialize) {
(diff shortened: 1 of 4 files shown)


dfinke:
Thanks for the PR! Wondering how to mock up a test for this, so we don't code it wrong down the road?

scrthq:
@dfinke I'm wondering if a simple parse test would pass? I'll try seeing what I can add :-)

scrthq:
Since this `finally` statement is buried within the `catch` statement, the error is only hit if you don't have the required dependencies when on Linux or macOS. Tried out a parser check, but no errors hit at all, so I don't think that's a valid test. We'll likely need to have a test that tries to import the module before those dependencies are installed on the CI container to confirm that the appropriate Warning is thrown and module import throws an error when dependencies are not found on *nix.

scrthq:
Test and additional focused stage added to Azure Pipeline definition, looks good! https://dougfinke.visualstudio.com/ImportExcel/_build/results?buildId=675&view=logs&j=aa21ab21-62eb-5dd0-25c4-258d984ac697&t=e4ee63a0-9882-57db-485d-edb2c576d271

dfinke:
Still under review

scrthq:
Heyo @dfinke - following up 🙂

dfinke:
Been trying to find time to test. :)

PR #837, opened 2020-05-04
Prevent install of Pester v5

dfinke:
Add `-MaximumVersion 4.99.99` on `Install-Module Pester`. In `yml` files and `ps1`.


dfinke:
@jhoneill , @ili101. Whenever it works, extra eyes for this would be great. Thanks.

jhoneill:
Agree. 
I think the stuff @ili101  did a while back meant we're aligned with 4's syntax and should find moving to 5 fairly easy.  Suffering a bit of brain fade, but I think the version on Azure DevOps is a bit out of date, I'm sure something I did need me to install 4 or a later version of 4.

PR #873, opened 2020-06-20
Standardising casing of parameter names

MikeyBronowski:
* standardise 'worksheet' as a single word, instead of 'workSheet' or 'WorkSheet'
* main parameter names starting with capital letter $WorksheetName
* supporting variable names starting with lower letter $worksheetName


dfinke:
@MikeyBronowski Thank you for the updates!

PR #922, opened 2020-09-17
Install latest Pester in CI

dfinke:


nlsdg:
Just to check the intended behaviour:
This doesn't alway update to the latest pester. It only updates if there isn't a version newer than 4.4.1.
So depending on you environment, it might not update to Pester v5.

dfinke:
Not sure what you mean? The `-MaxVersion` was removed, planning on using Pester v5 and latest versions as it updates.

nlsdg:
If you have Pester 4.4.1 or above, it will not update to Pester v5, due to the version check in line 249.
So you're not guaranteed use Pester v5, which might be important for breaking changes between Pester v4 and v5.
Like I said, this depends on the version already available (if any) on the system.
It might not be a problem, but is something to be aware of.

dfinke:
Thanks. Yeah, I'm moving this to GitHub actions. I'll double check that as I migrate.

dfinke:
In progress

PR #928, opened 2020-10-02
Add option to skip hidden/filtered rows

joelfish:
Add a switch parameter to skip hidden rows
Filter out hidden row indices after all other conditionals (header, startrow, dataonly, etc) have been processed for simplicity

### DIFF
--- a/Public/Import-Excel.ps1
+++ b/Public/Import-Excel.ps1
@@ -32,6 +32,7 @@
         [Alias('RightColumn')]
         [Int]$EndColumn  ,
         [Switch]$DataOnly,
+        [Switch]$SkipHiddenRows,
         [string[]]$AsText,
         [string[]]$AsDate,
         [ValidateNotNullOrEmpty()]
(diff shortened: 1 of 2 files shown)


dfinke:
Looks like a good feature. Please could you add a test for this so that we can be sure it works, and then also have a way to make it works in the future.

joelfish:
Thanks for the feedback! I've done tons of Powershell over the years, but
this is my first submission to someone else's work so I'm feeling my way
through it. Thanks for the patience. Any reference information you could
direct me to regarding the tests would be appreciated.

Also, thanks for this module. Love it and have used it for ages.

Joel Phillips

dfinke:
Great, take your time, no rush, and ask questions. Check out https://github.com/dfinke/ImportExcel/tree/master/__tests__. Use `Invoke-Pester` to run the tests. Do a `cd __tests__` in where you cloned the ImportExcel. Then add a ps1 file, follow the name pattern. Initially, you'll want to create an xlsx set up so you can test that your logic works.

joelfish:
Very cool thanks so much!

dfinke:
Looking good. Please add one more test, to make sure the records you get back  have the correct data from the rows that are not hidden.

dfinke:
Thanks, looks good. Will be merging and publishing soon. Thanks again for the contribution.

joelfish:
Awesome! Thanks so much for your help!

derricksmith:
Noticed this not merged.  Any plans to merge?

joelfish:
I still use this module daily and would still like this feature to be merged, but did my part (I think) and have just left it to the process.

derricksmith:
@dfinke , this is a valuable feature.  Can you merge?

PR #943, opened 2020-11-02
De-skipped some tests. Fixed a tiny bug in Add-Conditional formatting.

jhoneill:
As its says in the title.


dfinke:
I need to review

dfinke:
needs review

PR #951, opened 2020-11-19
Add Add-CellComment

mhokanson:


dfinke:
Much better. Tests? :-)

mhokanson:
I'm still working on learning Pester, but I'll see what I can do 👍

dfinke:
OK. It'd be a bit challenging. You'd need to create an xlsx on the fly to Pester's `testdrive`, do the `Add-CellComment` and then write code that tested the correct cell got set to the correct value.

Might add a `Get-CellComment` to make it easy to pull out a comment.

mhokanson:
Yep. I've got a Remove-CellComment and Get-CellComment in my local filesystem right now.

I am currently trying to decide how to handle addressing, as my Get-CellComment wants the column number, not letter, but I wrote Add-CellComment to accept the column letter. I want the addressing to be consistent across the three cmdlets.

mhokanson:
I just found your Get-ExcelColumnName cmdlet, so that'll make life easier for me.

mhokanson:
Looking at the Azure Pipeline logs, it looks like the tests failed because the new cmdlets weren't available for the tests that failed. How do I get the tests to load all three new cmdlets prior to running the tests?

dfinke:
Not sure. Need to dig when I get time. Looks like it might be pulling in an unexpected version of Pester. Will get back to you.

PR #954, opened 2020-11-23
Fixed the pester issues. (Locally and in AzDo pipelines)

jhoneill:
I have created mocks for get-service and get-ciminstance  on non-windows (load data exported with export-cliXML file) and fixed a couple of other broken bits. 
I've also changed the way parameters are passed into Pester and made sure the pester version shows up in the logs.


dfinke:
Thanks @jhoneill for working through this. Looks like not all the OSes are passing. Is this all due to moving to then new version of Pester?

jhoneill:
> Thanks @jhoneill for working through this. Looks like not all the OSes are passing. Is this all due to moving to then new version of Pester?

It seems to be the version pester is loading from the pipeline. I'm running my pipeline from /ci/pipeline.yml and that seems to get Pester 5 everywhere.  I think if yours runs off /azure-pipelines.yml you need to take out the `-MaximumVersion 4.99.99'` through out.  I think your pipeline always runs from your copy of the yml (otherwise a PR would run someone else's selections in your azure), so you need to change that at source.

dfinke:
I'll take a look. I had started this https://github.com/dfinke/ImportExcel/tree/FixPester. Was thinking of doing the move to Pester v5 and GitHub Actions and in one. I'll look at the AzDo, so there is a baseline for that.

PR #957, opened 2020-11-24
Add Comment functions and tests

mhokanson:
I think I finally figured out how to sync to your Master branch, which appears to now have some Pester fixes worked out.


mhokanson:
How can I set up my own build pipeline so I don't burn through all your Azure minutes trying to work this out?

mhokanson:
I ask about setting up my own pipeline because when I test locally here's my Pester results:

Tests Passed: 224, Failed: 0, Skipped: 0, Pending: 4, Inconclusive: 0

dfinke:
The module has been moved to Pester v5. Good question on your own pipeline. I believe you can set up a free Azure DevOps account for free and point it to your fork/PR. 

The https://github.com/dfinke/ImportExcel/blob/master/azure-pipelines.yml defines the pipeline.

PR #958, opened 2020-11-24
Add comment functions and tests

mhokanson:
I set up my own Azure pipelines and got all the tests to pass. Hopefully the same happens now.

### DIFF
--- a/ImportExcel.psd1
+++ b/ImportExcel.psd1
@@ -209,5 +209,4 @@ Check out the How To Videos https://www.youtube.com/watch?v=U3Ne_yX4tYo&list=PL5
 
     # Default prefix for commands exported from this module. Override the default prefix using Import-Module -Prefix.
     # DefaultCommandPrefix = ''
-
 }
--- a/ImportExcel.psd1
+++ b/ImportExcel.psd1
@@ -50,7 +50,8 @@ Check out the How To Videos https://www.youtube.com/watch?v=U3Ne_yX4tYo&list=PL5
         'DoChart',
         'Expand-NumberFormat',
         'Export-Excel',
-        'Export-ExcelSheet',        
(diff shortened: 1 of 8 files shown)


mhokanson:
Is there anything else you need for this? 

If you're okay with it I'd like to contribute more to this project. Would writing more Pester tests for other functions be helpful?

dfinke:
Thanks for the contribution. I have been looking at an upgrade to ImportExcel. I'm not sure if I will move forward with my upgrade yet. I wanted to avoid having to do the extra work of pulling over changes in this repo to another.

What else were you thinking of adding?

mhokanson:
At this point I was thinking I'd write Pester tests to help cover more functions in testing.

dfinke:
Thanks @mhokanson. Would love to have them. I'm currently looking at some options and directions to go with the ImportExcel module. It will take a while longer to investigate. I'll let you know when the choice is made. That way double work can be avoided.

Thanks again,

dfinke:
@mhokanson was looking at this. One challenge is I have a private repo where I'm experimenting with some things, and `git cherry-pick` is not my favorite things to do across repos. 

Wanted to put this forward, rather than using `column` `row`. I think keeping it consistent, following `Set-ExcelRange` would be good. Wanted to get your thoughts. Hold off on implementing.

mhokanson:
I like consistency, and if the project is moving towards using ranges instead of column/row then I think that makes sense. I'll see what I can do.

mhokanson:
How about this? What if I support both options by allowing a range _or_ column/row. 

I think I like that route better.

Thoughts?

dfinke:
The project will move forward, depends on if will it include the experiments I'm working out. 

Range, Column, and Row sound good. Range can be `A1` or `B2:C4` etc.

PR #959, opened 2020-11-24
Example: Removed a duplicate parameter Range/Address

MikeyBronowski:
As noted in the comments -Address is an alias for -Range

In current form it's giving an error:
```
Set-ExcelRange : Missing an argument for parameter 'Range'. Specify a parameter of type 'System.Object' and try again.
At line:1 char:16
+ Set-ExcelRange -Range -Address $sheet.Cells["E1:H1048576"]  -Horizont ...
+                ~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Set-ExcelRange], ParameterBindingException
    + FullyQualifiedErrorId : MissingArgument,Set-ExcelRange
```


dfinke:
Thanks Mikey!

PR #988, opened 2021-01-25
Add-ExcelName support for scope

mauricecote:
Excel creates Named Ranges using a Workbook scope, you must go out of your way to create in the WorkSheet scope.

I know this would mean a breaking change, but named ranges cannot be used in Data Validation scenarios, they are severely limited.

Obviously the switch can be inverted, but I feel it would be best to follow the same pattern as Excel


mauricecote:
I will fix the tests, and add new ones, but I'd like your thoughts before I start on that.

Rather then creating a breaking change, I could rather create a new Add-NamedRange cmdlet which would offer the choice and mark Add-ExcelRange as deprecated

dfinke:
Nice, thanks! Didn't know you could do worksheet scope. My first thought was to not make it the default, and avoid a breaking change. I like the suggestion. 

I'm in the middle of some things, so it'd be good if you didn't invest effort on all that just yet. 

I'll get back to this. Thanks again

PR #991, opened 2021-01-26
Example: fixed column name

MikeyBronowski:
The Quantity column is D, not F as in the example.


dfinke:
Thanks. `Unable to load shared library 'libmi' or one of its dependencies.` gets that error. I didn't that last time. More clues.

PR #998, opened 2021-02-12
Added example for exporting data to Excel for Pester 5.1

DarkLite1:
This example is compatible with the latest Pester. I left the old example for use with Pester 4.


dfinke:
Thank you for the example.

PR #1031, opened 2021-05-13
Fix UNC paths for ConvertFrom-ExcelSheet

sporkabob:
UNC paths broken with just Path property of Resolve-Path, looks like it was fixed elsewhere but overlooked here.


dfinke:
Thank you

PR #1057, opened 2021-07-30
Added compatibility for PS5 and lower

### LINKED ISSUE #1056: Read-Clipboard - Core only?
I noticed Read-Clipboard relies on $IsWindows variable to be populated which is only done in Core, is this intended?

DEberhardt:
Don't know whether the rest of the code would work, but as $IsWindows is not returning TRUE in PS v5, I thought to add $PSVersionTable.PSEdition -eq "Desktop"

fixes #1056

### DIFF
--- a/Public/Read-Clipboard.ps1
+++ b/Public/Read-Clipboard.ps1
@@ -22,7 +22,7 @@ function Read-Clipboard {
         $Header   
     )
     
-    if ($IsWindows) {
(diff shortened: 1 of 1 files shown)


dfinke:
Thank you! Looks like it returns `$false` on Linux. Going to refactor this a bit so I can add a pester test for checking where/what PS is running on

dfinke:
Thank you for PR, went with a different set of checks.

Issue #1064, opened 2021-08-12
Getting error when worksheet is named "Package"


jdenicola-pa:
In this example code:

```
$xlSourcefile = "c:\TEMP\ImportExcelExample.xlsx"
Remove-Item $xlSourcefile -ErrorAction Ignore

1..10 | Export-Excel $xlSourcefile -WorkSheetName "Package"
11..20 | Export-Excel $xlSourcefile -WorkSheetName "Package" -StartRow 15

$excel = Open-ExcelPackage $xlSourcefile 
Close-ExcelPackage -ExcelPackage $excel -Show
```

When I run this, it populates the spreadsheet with the data, but it gives me an error:

> Add-Member : Cannot add a member with the name "Package" because a member with that name already exists. To overwrite
> the member anyway, add the Force parameter to your command.
> At C:\Program Files\WindowsPowerShell\Modules\ImportExcel\7.2.2\Public\Open-ExcelPackage.ps1:38 char:17
> + ...             Add-Member -InputObject $pkgobj -MemberType ScriptPropert ...
> +                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>     + CategoryInfo          : InvalidOperation: (OfficeOpenXml.ExcelPackage:PSObject) [Add-Member], InvalidOperationEx
>    ception
>     + FullyQualifiedErrorId : MemberAlreadyExists,Microsoft.PowerShell.Commands.AddMemberCommand

If I change the WorkSheetName to something else - let's say "PackageX", it works without any errors.

Why does it care if the worksheet is named "Package"?

stahler:
Debugging your example, I see where "Package" is a property of the OfficeOpenXml.ExcelPackage object, thus the collision.

![image](https://user-images.githubusercontent.com/1991193/132776011-8d15bb11-6875-4c05-a2ff-f8c6e0b11794.png)

jhoneill:
This is something I put in a long time ago so that we refer to 
$excel.worksheetname as a shorthand  - I didn't trap a worksheet name being the same as a property or method.  It looks like you error is in public\Open-ExcelPackage.ps1 at line 38. So if you edit that and  make the `add-member xxxx`  command  at line 38 
` try {add-member xxx}  catch {} `
 that will give you a short term fix. When I get a chance I'll it properly for the next release

dfinke:
[...]

PR #1078, opened 2021-09-14
Add-Checks-And-Warnings-for-Add-Member-Open-Excel-Package

dfinke:
Addresses #1064


dfinke:
@jhoneill, did the try/catch. Figured this is enough for now, if someone adds a sheet with a `reserved` name, they'll get a warning if they use `Open-ExcelPackage`.

PR #1086, opened 2021-10-02
Escaping ' chars in field

wiebew:
The ConvertFrom-ExcelToSQLInsert did not escape ' characters and fails when a field has a ' character in it


dfinke:
Thanks for the PR. Please check/resolve the errors found in the checks that were run.

PR #1102, opened 2021-11-03
Add selective column import

muschebubusche:
Added a new parameter to Import-Excel which is called ImportColumns. It is used to define which columns of the ExcelPackage should be imported. If it is set StartColumn and EndColumn will be ignored.

Added several pester tests.


dfinke:
Interesting. Please could you post an example of it in action, so I can compare it to `StartColumn` `EndColumn`

muschebubusche:
> 
> 
> Thanks for the PR. Interesting. Please could you post an example of it in action, so I can compare it to `StartColumn` `EndColumn`

In our environment I use this for importing several Excel files with 20+ columns where only specific data is needed.
For example is there an HR file with a lot of columns but I only need B, T and Y.

image

image

Instead of using here `StartColumn` and `EndColumn` which would get me a lot more data than actually needed I can specify the columns with `ImportColumns` .

I hope this explains the usecase a bit.

dfinke:
@muschebubusche Got it. Import multiple non-contiguous columns. Makes sense. Thanks for the PR. I'll review.

dfinke:
@muschebubusche Nice feature. I'm reviewing another PR. Will merge this soon.
A few things:

- Could you add an ps1 example in the `Examples`  folder, create a `ImportColumns` folder. That way I can point folks to it 
- This be a great blog post, do you want to do one?
- I'll put this in the changelog. What links for you can I put there for you? Twitter, etc.

Thanks the addition!

dfinke:
## Bug?

I ran this in CodeSpaces, on Linux, I think PS 7.1.x. 

```powershell
Import-csv ./Examples/PassThru/sales.csv | Export-Excel ./sales.xlsx 
Import-Excel ./sales.xlsx -ImportColumns 1,3,2
```

### Expected:

```
Region UnitSold Item
------ -------- ----
South        54 Banana
West         74 Banana
West         26 Apple
```

### Actual

```
Region UnitSold UnitCost
------ -------- --------
South        54     0.46
West         74     0.56
West         26      0.7
East         38     0.26
East         71     0.69
East         35     0.55
```

muschebubusche:
@dfinke This bug is an expected behavior, because it cleans the Excel package at every iteration with this code:
`$ExcelPackage = Clear-ExcelPackage -Start $start -Count $count`

So the array needs to be in order. Maybe I'll add a check and display a warning if is not.

>Could you add a ps1 example in the Examples folder, create an ImportColumns folder. That way I can point folks to it

Yes I'll do!

>This be a great blog post, do you want to do one?

Why do you think so, I think it's a rather niche feature. But I can do one of course.

>I'll put this in the changelog. What links for you can I put there for you? Twitter, etc.

You can use my Github.

>I think it'd be cool to have -ImportColumns A,ZZ,B,YY3. May not be worth it, translating column "names" to indexes, etc.

I thought about the same thing but didn't look into it yet.

dfinke:
Thanks! Short blog posts on new features, just a good way to bring attention to it, to you for the contribution. Plus, long term it is a good way to point people to it, hit on it via search. Like an example PowerShell script showing it in action.

Just a nice to have, not a requirement :-)

muschebubusche:
@dfinke As an additional thought to adding a column translation - it wouldn't fit the function in its current state because if `ImportColumns` can translate it, then so should `StartColumn` and `EndColumn` aswell.

dfinke:
Thanks for catching that consistency.

dfinke:
Thanks for adding that. Just did a quick scan of the implementation. so you delete the columns not requested `.deletecolumn`? Interesting approach. When import-excel finishes, if down the road I fat finger import-excel and "save" the file, the user will lose data?

muschebubusche:
Yes you will loose data.

I thought about sorting the array if is not in order so there is no data loss but then I could also try to sort the result according to the original parameter `ImportColumns`. Unfortunately I didn't find an easy way to do it except using `$Worksheet.Cells[1..X].Copy` which isn't easy at all.

On the other hand I could display a warning, sort `ImportColumns` and give a result back not in the expected order but without data loss.

dfinke:
Just to be sure, the data loss I'm thinking of is, the code removes the columns from the spread sheet. If later, say I refactor Import-Excel, when the xlsx is closed, I forget to use the `-NoSave` switch, the xlsx is now updated and those columns are gone?

So a user calls Import-Excel, with -ImportColumns and the their original xlsx is now altered?

muschebubusche:
@dfinke I had some time to look into automatic column arrangement and added it to `ImportColumns`. So the topic about data loss should be solved with this addition.

>If later, say I refactor Import-Excel, when the xlsx is closed, I forget to use the -NoSave switch, the xlsx is now updated and those columns are gone?

If later `$ExcelPackage.Save()` is used instead of `$ExcelPackage.Dispose()`, then yes it will modify the original xlsx.

dfinke:
I like the feature. I have concerns with that potential side effect.

muschebubusche:
> I like the feature. I have concerns with that potential side effect. 

Thank you!
I'm not sure how to prevent that side effect. Maybe create a copy of `$ExcelPackage` can help.

dfinke:
Understood. I'm just not comfortable modifying the original data. At some point, because of an oversight, or a unit test not being in place, the package could be saved and the data lost.

 I like the feature and how you approached it. Still thinking through how to guard against accidentally losing data.

muschebubusche:
The new version of selective column import only manipulates `$Columns` in the function `Get-PropertyNames` and let Import-Excel do the rest. So the data will not be touched by my feature. 😄

dfinke:
Did a quick scan. You refactored away from deleting columns? I'll review when I can get to it.

Thanks for reworking this!

dfinke:
@muschebubusche Nice refactoring. The logic is cleaner, and "more capable". I've been thinking how to do this, did not come with anything. Good stuff, thank you.

I need to update the changelog.md etc, then publish. I will probably merge another PR first, so, hopefully soon.

dfinke:
@muschebubusche published

PR #1120, opened 2022-01-08
New-ConditionalText determine Range based on Column number of the Array you want to save to Excel

itiseverywhere:
This allows you to let Conditional Text 'move along' with columns in whatever you want to save to Excel.
This way you can re-arrange, add, remove columns, and the Conditional Text will stay with the right column, based on the column name.


dfinke:
Thank you for the PR! Please check the tests that are failing.

Issue #1041, opened 2021-06-17
Add System.Drawing.Image support


jhendricks123:
Thanks for such a great tool! It's really nice to be able to just chuck data at a cmdlet and see a spreadsheet pop out the other side 😎

I work for a company that works a lot with video and often times our customers would appreciate exporting snapshots of video to Excel to document camera positioning and image quality so I started to explore how to do that with your module and EPPlus today. I thought I'd make a feature request here as well as show how I managed to get it working.

I added another type check for `[System.Drawing.Image]` and used the worksheet.Drawings.AddPicture method to embed a picture into the sheet. I don't know if this is the best way to do it and I'm annoyed that I can't seem to constrain an image to a specific cell - either that's now how Excel works with images or I haven't worked out how to do it yet.

I also noticed that the column/row sizing use strange units and there's a very good chance the 375/500 and 17/120 ratios I worked out are specific to my environment/dpi/resolution/whatever. But these ratios are what enabled me, on my system, to produce the content in the screenshot below where the column and rows are resized to the dimensions of the image. I understand Excel has a maximum size for rows and columns so this doesn't take that into account yet either.

With the below `elseif` in place, I was able to pass a `[pscustomobject]` with a property containing a `[System.Drawing.Image]` object to `Export-Excel`

```powershell
elseif ($v -is [System.Drawing.Image]) {
    $picture = $ws.Drawings.AddPicture("$Name_$ColumnIndex_$row", $v)
    $picture.SetPosition($row - 1, 0, $ColumnIndex - 1, 0)
    if ($ws.Row($row).Height -lt $v.Height * (375/500)) {
        $ws.Row($row).Height = $v.Height * (375/500)
    }
    if ($ws.Column($ColumnIndex).Width -lt $v.Width * (17/120)) {
        $ws.Column($ColumnIndex).Width = $v.Width * (17/120)
    }
}
```

![2021-06-16 23_00_41-test xlsx - Excel](https://user-images.githubusercontent.com/7669972/122340148-c528e380-cef6-11eb-99d8-3b31ac4fb17e.png)

dfinke:
Nice work! Could you share how you set up the `[pscustomobject]` to pass in? Need to think through if this could be easily surfaced with the different checks for height and width. Plus supporting this may be a bit much.

An alternative way is to use `-PassThru` on `Export-Excel` and you can add the images to each row. You'd need to handle the row column.

Would like to see your code and will think about it.

Thanks, I like it.

[...]

PR #1133, opened 2022-02-02
Add Add-ExcelImage example

joshooaj:
Last year I opened #1041 to ask about support for Image objects in `Export-Excel`. Due to the complexity of working with images and image placement, and the relatively low demand compared to working with non-binary data sets, it turned out to be a better idea to add images as a "post-processing" step using the `-PassThru` switch with `Export-Excel` and then calling `$worksheet.Drawings.AddPicture($name, $image)`.

This PR adds an "AddImage" folder in Examples with a sample function in `Add-ExcelImage.ps1` and an `AddImage.ps1` script which demonstrates how to dot-source and use the function.


joshooaj (commented):
Thanks! I didn't realize `System.Drawing.Common` was not supported outside of Windows. Tested from Windows and Ubuntu via WSL and example works as expected in Windows, and throws as expected in Ubuntu.

PR #1145, opened 2022-03-17
Allow for escaping of single quote in data.

jamesmmueller:
If data in the Excel file contains single quotes, then the insert will bomb as it will be malformed. Quotes would need escaped. Not all languages will support the same escape method, so I recommend it being set by an optional parameter.

Example of what will fail:
Name
Fry's

INSERT INTO [Name] Values ( 'Fry's' )

Corrected for quotes (such as needing to escape a single quote with an additional single quote)

INSERT INTO [Name] Values ('Fry''s')

Example code:
$ConvertToSqlParams = @{TableName = ‘dbo.Names’
Path = ‘C:\temp\data.xlsx’
ConvertEmptyStringsToNull = $true
UseMSSQLSyntax = $true
SingleQuoteStyle = "''"
}

$SQLInsert = ConvertFrom-ExcelToSqlInsert @ConvertToSqlParams

### DIFF
--- a/Public/ConvertFrom-ExcelToSQLInsert.ps1
+++ b/Public/ConvertFrom-ExcelToSQLInsert.ps1
@@ -16,12 +16,15 @@ function ConvertFrom-ExcelToSQLInsert {
         [switch]$NoHeader,
         [switch]$DataOnly,
         [switch]$ConvertEmptyStringsToNull,
-        [switch]$UseMsSqlSyntax
+        [switch]$UseMsSqlSyntax,
+		[Parameter(Mandatory = $false)]
+		$SingleQuoteStyle
     )
 
     $null = $PSBoundParameters.Remove('TableName')
     $null = $PSBoundParameters.Remove('ConvertEmptyStringsToNull')
     $null = $PSBoundParameters.Remove('UseMsSqlSyntax')
+    $null = $PSBoundParameters.Remove('SingleQuoteStyle')
 
     $params = @{} + $PSBoundParameters
 
@@ -38,7 +41,12 @@ function ConvertFrom-ExcelToSQLInsert {
                 'NULL'
             }
             else {
-                "'" + $record.$propertyName + "'"
+                if ( $SingleQuoteStyle ) {
+					"'" + $record.$propertyName.ToString().Replace("'",${SingleQuoteStyle}) + "'" 
+					}
+				else {
+				 "'" + $record.$propertyName + "'"
+				}
             }
         }
         $targetValues = ($values -join ", ")


jamesmmueller:
Doug, is there anything more you need from me to help you troubleshoot this? I got the stale notice, and wanted to be proactive in helping. I plan on using your ImportExcel for automating some file imports from a vendor, but would prefer not to fork it to resolve the quoting issue.

dfinke:
Thanks. I know it doesn't look like a big or complicated change, from a quick glance it looked good. I have not had the time to check it out in terms of testing. Plus, iirc (and maybe not), I don't have a lot of tests for this, so I am more hesitant about it breaking things

jamesmmueller:
@dfinke, I understand fear of breaking something. This is what I did to test the error to track down that it was occurring before making changes:

**Steps in SQL Server Management Studio**:
CREATE TABLE TestMyFile ( TestColumn VARCHAR(12) )

**Steps in Excel**
Excel File called "C:\Temp\TestQuotes1.xlsx" 
A1  TestColumn
B1  Test 

Excel File called "C:\Temp\TestQuotes2.xlsx"
A1  TestColumn
B1  Test's 

**Steps in PowerShell**
Import-Module SQLServer

$ConvertToSqlParams1 = @{TableName = ‘dbo.TestMyFile’
  Path = ‘C:\temp\TestQuotes1.xlsx’
  ConvertEmptyStringsToNull = $True
  UseMSSQLSyntax = $True
}

$SQLInsert1 = ConvertFrom-ExcelToSqlInsert @ConvertToSqlParams1
Invoke-Sqlcmd -Query "${SQLInsert1}" -ServerInstance "Server" -Database "Database"

#This should have inserted a row. In SQL Server use  "SELECT * FROM TestMyFile" to see the values

$ConvertToSqlParams2 = @{TableName = ‘dbo.TestMyFile’
  Path = ‘C:\temp\TestQuotes2.xlsx’
  ConvertEmptyStringsToNull = $True
  UseMSSQLSyntax = $True
}

$SQLInsert2 = ConvertFrom-ExcelToSqlInsert @ConvertToSqlParams2
Invoke-Sqlcmd -Query "${SQLInsert2}" -ServerInstance "Server" -Database "Database"

_Invoke-Sqlcmd:
Line |
   1 |  Invoke-Sqlcmd -Query "${SQLInsert2}" -ServerInstance "Server" -Database "Database"
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Unclosed quotation mark after the character string._
 
Write-Host $SQLInsert2
_INSERT INTO dbo.TestMyFile ([TestColumn]) Values('Test's');_
   
     
**When done do this in SQL Server**
`DROP TABLE TestMyFile`

SQL Server is interpreting the value as 'Test' and then trying to parse s'); after it and failing. Adding the SingleQuote style changes the value to have all quotes become two single quotes which escapes the quote. 

If you can give me a good workaround to replace all single quotes in the values of an Excel file with two single quotes, I'd do that instead. I'm not well versed in ImportExcel yet. This was my first use case for it.

dfinke:
Thanks for the PR! Looks good, thanks for the update. After looking closer, the added param and conditional ought to be fine.

dfinke:
Thank you again. v7.4.2 is published. https://github.com/dfinke/ImportExcel/releases

PR #1159, opened 2022-04-10
Add Read-Excel function

dfinke:


dfinke:
Add this to `Import-Excel` in https://github.com/dfinke/ImportExcel/releases/tag/v7.5.0

Issue #1188, opened 2022-06-01
Two errors thrown when a worksheet cannot be found


DarkLite1:
When using `Import-Excel' to retrieve the content of a specific worksheet that is not available there are two errors thrown while only one is expected.

### Example

```powershell
$params = @{
    Path          = New-TemporaryFile | 
    Rename-Item -NewName { $_ -replace 'tmp$', 'xlsx' } -PassThru
    WorksheetName = 'Sheet1'
}
@(
    [PSCustomObject]@{'Fruit' = 'Banana' }
    [PSCustomObject]@{'Fruit' = 'Kiwi' }
) | Export-Excel @params

Import-Excel -Path $params.Path -WorksheetName 'nonExisting'
```
### Errors
![image](https://user-images.githubusercontent.com/10828570/171344075-809642e0-c17a-4dc6-aec0-a14fb921dd55.png)

### Expected

Only one error:

> Failed importing the Excel workbook 'C:\Users\bgijbels\AppData\Local\Temp\3\tmp3039.xlsx' with worksheet 'nonExisting': Worksheet 'nonExisting' not found, the workbook only contains the worksheets 'Sheet1'. If you orkbook only contains the worksheets 'Sheet1'. If you only wish to select the first worksheet, please remove the '-WorksheetName' parameter.

DarkLite1:
That's what we already do. In normal circumstances a terminating error of a function only throws a single error and does not pollute the error object with multiple errors originating from the same function. 

So in a sense, you're right, try/catch is the way. But it should be done in the function and clean-up the old produced errors into one really easily digestible error for the end user. Hence, that's why I already made the PR to fix this.

PR #1190, opened 2022-06-01
Fix for #1188

DarkLite1:
- Fix for #1188 : cannot index into 0 error
- Improved error handling in "Import-Excel" to throw consistent error messages (tests updated too)
- Add a test for producing only one error when "Import-Excel" fails instead of multiple
- Some small typo's fixed

### DIFF
--- a/Public/Import-Excel.ps1
+++ b/Public/Import-Excel.ps1
@@ -111,7 +111,7 @@
             try {
                 if ($path) {
                     $extension = [System.IO.Path]::GetExtension($Path)
-                    if ($extension -notmatch '.xlsx$|.xlsm$') {
+                    if ($extension -notMatch '.xlsx$|.xlsm$') {
                         throw "Extension type '$($extension)' not supported"
                     }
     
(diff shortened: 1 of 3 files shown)


DarkLite1:
@dfinke is there a reason why this is not merged? Did I miss something?

dfinke:
@DarkLite1 I have not had time to look at it

DarkLite1:
@dfinke can you have a look if this can be implemented? It's no structural change and shouldn't break anything. The only thing it does is report a single error in case there is one instead of also throwing the internal errors separately

dfinke:
Sorry. Took a cursory look. My first thought was folks may be dependent on the error message as is

DarkLite1:
That would seem weird... Being dependent on the error message "Cannot index into a null array":
image

dfinke:
If they try/catch on that. Wouldn't your change break them?

DarkLite1:
It seems logical that if people expect 2 errors and there's only 1 that their expectations are not met. But it seems even more logical to handle the errors inside the function and throw a single consistent error where it's clear that this error is generated within the function, instead of polluting the `$error` array with internal errors coming from `Import-Excel`. This brings even more confusion to the table... 

IMHO no one is relying on an error count in this case. So I see no reason not to include it and have clean error handling.

DarkLite1:
Just for clarity, the `try/catch` clause would still work (see the issue I posted).

PR #1200, opened 2022-06-17
Update to Get-HTMLTable

jhoneill:
When you do `Invoke-WebRequest` in Windows PowerShell it returns a `Microsoft.PowerShell.Commands.HtmlWebResponseObject`  which has a `.ParsedHTML` property. 

Because of the changes from .NET Framework to .NET core in PowerShell 6+ the object is a `Microsoft.PowerShell.Commands.BasicHtmlWebResponseObject`  which doesn't have the property, so things break. 

The generally accepted solution is to use someone else's HTML Parser, and Justin Grote has a module  `PowerHTML` which wraps one. So I've modified Get-HTML Table

```
If version >= 6 and Justins command not found  then throw
download the html as before
if Version <   6 then  process it the way it always did   
else   use justin's command to do it. 
```

The new bit will take a bit of xpath to find the table e.g.
 ```
Get-HtmlTable -Url "https://www.baseball-reference.com/leagues/majors/2022.shtml" -TableIndex  "@id='teams_standard_batting'"
```

I've tried to get it converting text to numbers, and being fairly smart about header rows but I've only tried a small number of pages.  

There is also  an update to the import by columns example which I committed but never PR'd


dfinke:
`Import-ByColumns` 👍 - looks like a `transpose`, am I thinking about that correctly?

jhoneill:
> `Import-ByColumns` 👍 - looks like a `transpose`, am I thinking about that correctly?

Yes :-)  Did it ages ago to show it could be done.

PR #1208, opened 2022-07-05
Kkazala/exportformatting

kkazala:
@dfinke  Would you be interested in one new, and one updated example?
I updated the **Highlight-DiffCells.ps1** to format both compared cells, and the whole row:
image

And I added an example (**GetConditionalFormatting.ps1**) that exports conditional formatting rules in a "ready to use" format, so if someone has better grasp on Excel than the `Add-ConditionalFormatting`, they won't need to open an issue with you 
image

And a question: would you consider mentioning the examples on https://github.com/dfinke/ImportExcel? I'm a bit ashamed to admit, but I had no idea they are there until you asked me to contribute :see_no_evil:


dfinke:
@kkazala Thank you for the PR and suggestion about adding the examples to the readme. 

Is it possible, I like the `GetConditionalFormatting.ps1`, two things.
1.  Re-work that as a `function Get-ConditionalFormatting` in the ps1 you have
2.  Add a an xlsx so when I run your cool ps1, I can see this in action.

I may add this to this to the core library.

Thanks again.

kkazala:
@dfinke cool, will do

And... you just gave me an idea. How about, instead of an example, we could make it into a function? 
I only had problem with a formatting color, since Excel is using theme numbers and not colors
 
If you still would like it to be an example, could you suggest a folder the excel? I don't want to disorganize the project:)

dfinke:
Thanks. I like the idea of an export formatting function. I usually create examples and so I can try before I promote it to the core. That way it fleshes out issues, edge cases etc. Helps for writing unit tests.

For the excel file for the example? You can put it next to the example ps1 file. If I understand your question.

kkazala:
@dfinke I added the Excel as requested, the file name is GetConditionalFormatting.xlsx to make sure the excel and the script are next to each other

Please let me know if there's anything else

dfinke:
Please double check this line. The `$worksheetName` is blank when it prints. Plus, don't need the `Write-Host`.

I like the idea behind this. Dump Excel constructs as PowerShell code.

```powershell
Write-Host "Add-ConditionalFormatting -Worksheet `$excel[""$worksheetName""]  -Range '$($_.Address)'  -ConditionValue '$($_.Formula)' -RuleType $($_.Type) "
```

kkazala:
@dfinke Sorry about that, I was a bit in a rush. Fixed now.

Issue #1209, opened 2022-07-05
File name with Wildcards in Open-ExcelPackage


Alphagorn:
The problem arises with files that contain, for example, [Test][First]File123.xlsx. Especialy with the brackets.
In your Open-ExcelPackage you use a test-path -Path, which prevents opening these files. 
It would be a good idea to use a test-path -LiteralPath 
I modified the function and I included it in one of my modules which allowed me to troubleshoot.

Alphagorn:
Hello Doug,

I'm using your module and the Open-Excelpackage function with the Power BI API.

I save Power BI spaces and some file names are of the type "[PriorIT][POGS] Solicitations Infogérant.pbix".
However the [] are seen by the Test-Path function as wildcards and returns false with the -Path parameter.
If you use the -LiteralPath parameter the file name is interpreted as it is written and detected.

In your Open-ExcelPackage function, you use the Test-Path with -Path and not with the -LiteralPath parameter.
The error results in the fact that the Export-Excel function correctly created the file of type [xxx][xxx] xxx.pbix
but it cannot be opened by Open-ExcelPackage because it's not detected by the Test-Path.

I changed the function for my needs, but I think it is better to give the possibility 
of having the parameter -Path or -LiteralPath, although in the opening of a specific file we never use wildcards normally

So I changed your function like this:

2 lines from you :
if ($Create -and -not (Test-Path **-Path** $path)) {
        #Create the directory if required.
by:
if ($Create -and -not (Test-Path **-LiteralPath** $path)) {
        #Create the directory if required.

AND

elseif (Test-Path **-Path** $path) {
        if ($Password) { $pkgobj = New-Object -TypeName OfficeOpenXml.ExcelPackage -ArgumentList $Path , $Password }

by:
elseif (Test-Path **-LiteralPath** $path) {
        if ($Password) { $pkgobj = New-Object -TypeName OfficeOpenXml.ExcelPackage -ArgumentList $Path , $Password }


#---------------------------------------------------------
function **Open-ExcelPackage2** {
    [CmdLetBinding()]
    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingPlainTextForPassword", "")]
    [OutputType([OfficeOpenXml.ExcelPackage])]
    param(
        #The path to the file to open.
        [Parameter(Mandatory = $true)]$Path,
        #If specified, any running instances of Excel will be terminated before opening the file.
        [switch]$KillExcel,
[...]

PR #1212, opened 2022-07-08
Investigate LiteralPath for - File name with Wildcards in Open-ExcelP…

dfinke:
…ackage #1209


dfinke:
Not going to merge. This "breaks" `Import-Excel da*.xlsx`. Most likely all functions using `Test-Path`, those functions would need to handle both `Path` and `LiteralPath` to get it right.

Issue #487, opened 2018-11-02
Pivot Table "Magic" Pivot Date Row


dprice:
Here's some code that simulates reading a log4net log, adding timestamp 30 second bucket data, and launching Excel with a pivot table on the timestamp bucket count:
```
$results = ConvertFrom-Csv @"
Timestamp,Tenant
10/29/2018 3:00:00.123,1
10/29/2018 3:00:10.456,1
10/29/2018 3:01:20.389,1
10/29/2018 3:00:30.222,1
10/29/2018 3:00:40.143,1
10/29/2018 3:00:50.809,1
10/29/2018 3:01:00.193,1
10/29/2018 3:01:10.555,1
10/29/2018 3:01:20.739,1
10/29/2018 3:01:30.912,1
10/29/2018 3:01:40.989,1
10/29/2018 3:01:50.545,1
10/29/2018 3:02:00.999,1
"@ | % {
    $dt = Get-Date($_.Timestamp)
    $data = [ordered]@{
        Timestamp = $dt
        Tenant = $_.Tenant
        Bucket = $dt.AddSeconds(-$dt.Second % 30)
    }
    New-Object -TypeName psobject -Property $data
}

$excelParams = @{
    Path              = 'd:\temp\pivottest.xlsx'
    WorkSheetname	  = "Log Data"
    IncludePivotTable = $true
    PivotTableName    = 'Timestamp Buckets'
    PivotRows         = @('Timestamp', 'Tenant')
    PivotData         = @{'Bucket' = 'count'}
    Activate          = $true
    AutoSize          = $true
    AutoFilter        = $true
    Show              = $true
}
Remove-Item $excelParams.Path -ErrorAction Ignore
$results | Export-Excel @excelParams
```
It works great:
![image](https://user-images.githubusercontent.com/84499/47938502-a64abf80-deba-11e8-856a-8e9e2b7fc85f.png)

But when you uncheck `Timestamp` and recheck it, Excel magically creates `Hours` and `Minutes` columns, which is what I really want:
![image](https://user-images.githubusercontent.com/84499/47938632-16594580-debb-11e8-9c7f-54f98b3811a4.png)

Any idea on how to get these columns created from the script?

jhoneill:
@dprice  
Support for this was added in this weekend's new release 

```
[...]

PR #1219, opened 2022-07-14
Pivot table updates

jhoneill:
As discussed in #487


dfinke:
@jhoneill Is it worth adding tests for `GroupNumericColumn` and `GroupDateColumn`?

jhoneill:
> @jhoneill Is it worth adding tests for `GroupNumericColumn` and `GroupDateColumn`?

Not sure of the value, but strictly speak, yes, its a bit of functionality, so it should be tested. We can probably take the row tests and change them to column tests.

Issue #1231, opened 2022-08-18
Conditional Formatting 


g-pearl:
I am trying to follow along with:
https://dfinke.github.io/powershell/2020/05/02/PowerShell-Excel-and-Conditional-Formatting.html

But I need to highlight the bottom 1 item not the top 10.

From the above website I had to adjust the Table name and Worksheet name to Rawdata from StoresTop10Sales 
Here is my updated command.
Add-ConditionalFormatting -Worksheet $xl.Rawdata -Address $xl.Rawdata .dimension.address -RuleType Top -ForegroundColor white -BackgroundColor green -ConditionValue 10

I get the output to match the page but when I change the statement from top 10 to bottom 1 excel opens and it highlights the bottom 10 not bottom 1.

g-pearl:
I think I have found the issue.

I would be glad to fix it but I have never done a pull request - if that is even the right thing.

on Line 121 of Add-conditionalformatting there is the following

$RuleType -match "Top|Botom" 

If I change to: (+ "t")
$RuleType -match "Top|Bottom" 

The code works

dfinke:
Sorry missed this. Interesting could you post the data, script and xlsx that shows the problem. And then how you solved it.

Yes, a PR would be the way to do it. I need to look to see if there are pester tests for this. If not, they need to be added first, and then additional ones around the issue found.

g-pearl:
I completed a pull request earlier today - Please tell me if I did this wrong (First time)

dfinke:
Thank you! You did it correctly. Thanks again for finding and fixing.

PR #1233, opened 2022-08-24
Update Bottom Spelling Error

g-pearl:
This is to correct an error in Conditional formatting that was preventing "Bottom" from working.


dfinke:
Ah, got it! Excellent catch, thanks for the PR!

https://github.com/dfinke/ImportExcel/issues/1231

PR #1251, opened 2022-09-15
Update Add-Worksheet.ps1 - 'worsheet' to 'worksheet' in warnings

JeremiahTheFirst:
Due to some foolishness on my part, I saw this warning several times, so the 'worsheet' typo started to get to me.


dfinke:
Thanks for the "k". It's great when the community contributes like this.

PR #1254, opened 2022-09-21
Fixed typo of 'pararameter' to 'parameter'

WilsonStewart:
Noticed while using the Set-ExcelRange/Set-Format function.


dfinke:
Thank you! Very helpful

Issue #1184, opened 2022-05-11
NET Core, EPPlus and -Password


sztaylorakgov:
I see the same problem I'm encountering (ImportExcel 7.5.2) noted in #637, per comment: https://github.com/dfinke/ImportExcel/issues/637#issuecomment-592636559

@dfinke indicates in one comment in that issue from July 24, 2019:
> I checked a couple months back on EPPlus to see if they were further along on their port to dotnet core. Don't know about the crypto support. I didn't see much check in activity then.

I'm out of my depth on how to contribute a fix, but I did find this sample from EPPlus released after @dfinke comment that deals with encryption: https://github.com/EPPlusSoftware/EPPlus.Sample.NetCore/blob/master/19-EncryptionAndProtection/EncryptionAndProtectionSample.cs?msclkid=ad003fd1d0ce11ec99e9f81da0377d80

Again, I'm out of my depth, but that sample code makes me think that EPPlus has dealt with the dotnet core encryption port?

The current version of ImportExcel (7.5.2) fails on line 124 of Import-Excel.ps1.  It's probably no small task to just update the EPPlus.dll that is 3 years old in the repo - I expect there's potential significant regression testing against an update like that, but I wanted to raise this old issue in case the EPPlus updates to support encryption were not known.

Any chance this support for password encrypted spreadsheets and PowerShell Core can be revisited now that EPPlus seems to have a released version that supports it?

dfinke:
Thanks, no plans to move to that version. That EPPlus version is their new version. No real plans to move to it. It requires a paid license for each person that uses it, I have no influence over it. Also, the upgrade has a lot of breaking changes.

sztaylorakgov:
Good clarification.  I thought in reading EPPlus licensing terms that they had support for develop/distribute scenarios, i.e., developer pays for a license but can distribute.  But either way, breaking changes is a big deal.

I have a scenario where I don't control the workbook source and the workbook comes password encrypted.  Do you have any recommended work-arounds for continuing to use Import-Excel or other options for those scenarios?  I suppose the answer might be to just keep using PowerShell 5.1?

dfinke:
I talked with the devs of EPPlus. Lots of grey areas. Plus, not sure I want to shell out annual/one time $$.

I don't know of work arounds for this. If EPPlus isn't able to handle the decryption, I think 5.1 is the way forward. 

There may be someone who found a workaround with older .NET and the older EPPlus, check Stackoverflow.

PR #1260, opened 2022-09-29
EPPlus core (Password)

ili101:
- PowerShell before 5 continue to use "net40", PowerShell Core use "netstandard2.0", fixes `NET Core, EPPlus and -Password` [#1184](https://github.com/dfinke/ImportExcel/issues/1184)

Maybe need more testing as dll are loaded differently and new dependencies dlls to see if it brake something?

Note that I tried at first EPPlus 4.5.3.3 but 2 tests failed (only on PS 5.1), not sure if problem with tests or EPPlus, anyway I left the version as it was (4.5.3.2)


ili101:
Enabled tests, Windows Core password tests failed on Azure, run it again passed 🤷‍♂️

ili101:
Didn't dig into it, those are the 2 errors (only on 5.1) https://github.com/ili101/ImportExcel/runs/8623601775

dfinke:
@ili101 Thanks for putting this together. IIRC the last version of EPPlus introduced many issues. Need to come up with a broad UAT and burn in approach. Not sure if that'll happen.

PR #1273, opened 2022-10-14
Set Validate Range for rows to max rows available

steve-daedilus:
$StartRow parameter had ValidateRange(1,9999) which was limiting the start row to a maximum of 9999. There's no good reason this cannot be any row in the spreadsheet, of which the maximum is 1048576 in modern instances of Excel.


dfinke:
Thank you for the PR. Good job!

PR #1300, opened 2022-11-04
Add TotalSettings parameter to Export-Excel

thkn-hofa:
Adding the parameter TotalSettings to the function 'Export-Excel' allows for an easy implementation of 'TotalSettings' in stead of having to do manual work.


thkn-hofa:
Hey man, this is my first time contributing to a public project and I'm not a developer by education or profession. I'd love to provide the necessary but you'll need to point me in the right direction. Do I need to update the export-excel.tests.ps1?

dfinke:
No problem. Copy the `export-excel.tests.ps1` to `export-excel-TotalSettings.tests.ps1`. Delete everything but leave a single test where you can test the `TotalSettings`.

You need to have `Pester` installed so you can run the tests locally. 
- `cd __tests__`
- `Invoke-Pester`

They should all pass, there are some that show up skipped, that is a good run. Then, check in the code.

Once there are tests I can review.

Feel free to ask questions.

Thanks again!

thkn-hofa:
I think the tests are more than adequate. Should I put my 'Context' part in the Export-Excel.Tests.ps1 and remove my Export-Excel-TotalSettings.Tests.ps1 ?

__tests__/Export-Excel-TotalSettings.Tests.ps1
```diff
@@ -0,0 +1,71 @@
+﻿#Requires -Modules @{ ModuleName="Pester"; ModuleVersion="4.0.0" }
```
thkn-hofa, line 1: My Windows 10 machine had Pester 3.4.0 **pre-installed** which was unable to perform the tests. Turns out you need at least Pester 4.

dfinke (commented):
I'd prefer not to pin the Pester version

thkn-hofa:
The way I did it only defines the minimum version, it's not pinned on 4.0.0:

ModuleVersion - Specifies a minimum acceptable version of the module.
MaximumVersion - Specifies the maximum acceptable version of the module.
RequiredVersion - Specifies an exact, required version of the module. This can't be used with the other Version keys.

[Source](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_requires?view=powershell-5.1) 

I'd expect a parameter like MinimumVersion, too, but I'm running 5.3.0 or smth like that and everything works.

The versions before 4.0.0 don't work with the syntax the current tests are written in. 

I'll be happy to comply if you're still not convinced. If you are, I'm finished with my changes.

dfinke:
@thkn-hofa this is looking great, thanks for dotting the i's and crossing the t's. Could you add one more part. What is in the tests, could you add that as an example in the Example folder. - Thanks

dfinke:
@thkn-hofa Thanks for the feature! Very useful. Good to see how the `Custom` part played out, always good to learn something new.

thkn-hofa:
@dfinke With so much pleasure

Issue #1299, opened 2022-11-03
7.4.2 vs newer versions processing time


dfinke:
### Discussed in https://github.com/dfinke/ImportExcel/discussions/1279

<div type='discussions-op-text'>

<sup>Originally posted by **LeonidR2** October 19, 2022</sup>
I was tasked to import Excel file with 3 columns and about 100,000 records.  

$FileContentXLS = Import-Excel -Path $FileName -WorksheetName $WorkSheetName

Version 7.4.2 performs task in 20 seconds, while every new version completes task in 15 minutes. 

Would anyone know reason for such discrepancy? Can same optimization be applied to newer versions? </div>

PR #1301, opened 2022-11-06
Initial timing tests. 7.4.2 is the benchmark  7.4.2 vs newer versions…

dfinke:
Identify and fix processing time for Import-Excel #1299


dfinke:
review

PR #1324, opened 2022-11-23
Update to parse date strings as OLE Date

HunterSch-713:
Ran into some issues while importing a fairly large Excel document and the date time formatting / column definitions within it not being consistent, this is the fix that worked to check and convert strings that normally "[DateTime]::FromOADate()" couldn't parse because well they're strings and not OLE Dates. It seems like a good addition to beef up the date parsing from imported documents to lessen the burden of defining column types in Excel but that is only my own opinion.


dfinke (commented):
Thanks for the PR. I'd need unit tests for this. I didn't check, not sure if there are tests for the code before. 

Plus, I introduced a performance issue in `Import-Excel` that I am working through. 

This wouldn't be merged till that work is done and this code is vetted.

PR #1351, opened 2022-12-21
Example

oliwex:
Add one test file as an example in Examples\CommunityContributions


oliwex:
I will be gglad if you approve it. Just for my check if i can code in PowerShell :)

dfinke:
Looks good. Already have: https://github.com/dfinke/ImportExcel/blob/master/Public/Get-ExcelFileSummary.ps1

PR #1365, opened 2023-01-04
Description to Add-ConditionalFormatting and Add-ExcelChart

oliwex:
Add description(descript parameters, add few examples as Get-Help) to check and review by community


dfinke:
Thanks won't do this now. Will revisit.

Issue #1429, opened 2023-03-30
Export-Excel - Add parameter to set custom formats for datetime and timestamp columns


danubie:
Hello Doug,
when using Set-ExcelColumn/Row/Range param -NumberFormat can be used to set custom formats.

I would like add the option to set custom formats in `Export-Excel` as well, defaulting to the currently used formats.
```
[string]$DateTimeFormat = 'm/d/yy h:mm'
[string]$TimespanFormat = '[h]:mm:ss'
```
or
```
[string]$DateTimeFormat = (Expand-NumberFormat 'Date-Time')
[string]$TimespanFormat = '[h]:mm:ss'
```
The later would be compatible to be used for exporting PSCustomObject and DataTable (where 'Date-Time' is used to set the datetime format). So both cases could then be handled the same way.
What do you think about this enhancement and getting a PR for it to implement?

dfinke:
Interesting. Could you say a bit more about it, example usage? How do you see `-NumberFormat` changing?

Thank you

scriptingstudio:
I like the idea.
line 242 in Export-Excel.ps1
```powershell
if (-not $DateTimeFormat) {$DateTimeFormat = 'Date-Time'}
foreach ($c in $InputObject.Columns.where({ $_.datatype -eq [datetime] })) {
	Set-ExcelColumn -Worksheet $ws -Column ($c.Ordinal + $StartColumn) -NumberFormat $DateTimeFormat
}
if (-not $TimespanFormat) {$TimespanFormat = '[h]:mm:ss'}
foreach ($c in $InputObject.Columns.where({ $_.datatype -eq [timespan] })) {
	Set-ExcelColumn -Worksheet $ws -Column ($c.Ordinal + $StartColumn) -NumberFormat $TimespanFormat
}
```

danubie:
There's not much to be done. There are two places to be adopted.
```
Set-ExcelColumn -Worksheet $ws -Column ($c.Ordinal + $StartColumn) -NumberFormat 'Date-Time'
Set-ExcelColumn -Worksheet $ws -Column ($c.Ordinal + $StartColumn) -NumberFormat '[h]:mm:ss'
and
$ws.Cells[$row, $ColumnIndex].Style.Numberformat.Format = $DateTimeFormat
$ws.Cells[$row, $ColumnIndex].Style.Numberformat.Format = $TimespanFormat
```
The 2nd I've already done as a PoC to see if the proposed params would work and that the result is as expected.
[...]

PR #1430, opened 2023-03-30
Add-parameters-for-Datetime-and-Timespan-columns

danubie:
Implementing custom datetime and timespan formats for Export-Excel (#1429)

Adding a new param -CustomFormats <hashtable>


danubie:
Sure, there want be a fix in the library for reading back datetime values with special formatting correctly. So, should we keep this enhancement or drop it? 
On the other hand, this would be the same as if the unsupported formats are used "by hand". For this a solution could be providing a "Repair-" function (e.g. in the examples folder).

PR #1522, opened 2023-10-11
Add missing parameter aliases to align with caller/callee

kyllath:
Fix issue where none of 'HeaderRow', 'TopRow' and 'StartRow' work in ConvertFrom-ExcelToSQLInsert, preventing you from skipping rows.

Issue is that aliases are missing from the parameter list of ConvertFrom-ExcelData. 
ConvertFrom-ExcelToSQLInsert accepts 'HeaderRow', 'TopRow' and 'StartRow', but passes 'StartRow', while  ConvertFrom-ExcelData only accepts 'HeaderRow'

FIX: Alter ConvertFrom-ExcelData to have the same aliases for the parameter as both its caller, ConvertFrom-ExcelToSQLInsert and its callee, Import-Excel. All three now have:
        [Alias('HeaderRow', 'TopRow')]
        [Int]$StartRow = 1,

### DIFF
--- a/ImportExcel.psd1
+++ b/ImportExcel.psd1
@@ -6,7 +6,7 @@
     RootModule         = 'ImportExcel.psm1'
 
     # Version number of this module.
-    ModuleVersion      = '7.8.5'
+    ModuleVersion      = '7.8.6'
 
     # ID used to uniquely identify this module
     GUID               = '60dd4136-feff-401a-ba27-a84458c57ede'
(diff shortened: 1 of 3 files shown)


dfinke:
Hmm. Feels like TopRow should be added as a param not an alias. It's been a long time since I've benn in this part of the code.

kyllath:
The reason I chose this approach is that ConvertFrom-ExcelToSQLInsert always calls ConvertFrom-ExcelData, which in turn calls Import-Excel. Both ConvertFrom-ExcelToSQLInsert and Import-Excel have StartRow as the param with HeaderRow and TopRow as aliases, so this way just makes all three consistent.

dfinke:
@kyllath Thank you for the update!

Issue #1531, opened 2023-10-25
Refactor Question - Improving ImportExcel module import time


joshooaj:
Hi @dfinke,

Are you open to a PR to change the way the module is "built" for testing and publishing to reduce the amount of time it takes to import the module?

On my Windows 11 laptop it takes ~7.5 seconds to import the module in PowerShell 5.1 as published to PSGallery. The bulk of that time is spent on dot-sourcing the .PS1 files in the `Private`, `Public`, `Charting`, `InferData`, and `Pivot` subfolders. The next longest import step is importing plot.ps1 which only takes ~98 milliseconds.

In a quick and dirty test where I take all the contents of those subfolders and embed them in the .PSM1 file, the import time drops to only 362 milliseconds. The issue is less noticeable in PowerShell 7 where the current import time is only 1.1 seconds, but "compiling" all the functions into the .PSM1 file still reduces the import time by 20x, down to 53 milliseconds.

I haven't started yet but the idea would be to keep the current project structure, and change the build/CI process so that the version of the module being tested and published is "compiled" by inserting as much as possible into the PSM1 file beforehand.

I use this pattern another module where there are a total of 197 files in the Classes, Public, and Private folders with each function residing in its own file. The published version of the PSM1 file has over 15k lines and imports in ~3.6 seconds.

dfinke:
Thanks for the question. Have thought about that for a while. I'd be open to it.

I'm running Win 11 both version of PS import using -force in ~2 sec.

I have not thought through the scenarios where bugs could arise or debugging efforts.

joshooaj:
Okay, I'll get started! What's your current CI? I see references to Azure DevOps, AppVeyor, and GitHub Actions and I'm not sure if DevOps/AppVeyor are still being used or if those references are stale.

dfinke:
Thanks, really appreaciate it! GitHub Actions is preferred. Az is still enabled.

Note: I'm not sure how I can get this into enough folks hands or forewarn users of this sizable and potential impact. Also, I an not confident of the test coverage to uncover issues.

Additionally. I have other commitments. I may not get to review this for some time.

Given these risks, it could be months before I decide against it or merge the PR.

joshooaj:
Thanks for the notice - I can understand that. If you end up merging any of it, maybe it marks a major version increment to signal that there _might_ be breaking changes (even though the goal is that there aren't).

[...]

PR #1535, opened 2023-11-08
Refactor of ImportExcel project

joshooaj:
This PR started as an effort to reduce the time to import the module by taking the PS1 files that were previously dot-sourced by the PSM1 during import and "compiling" them into the PSM1 file before testing or publishing. This results in a faster import time on all platforms, but it's most noticeable on Windows and PowerShell 5.1 as discussed in #1531.

The refactor loosely applies ideas I learned from the PowerShell community over the last few years and especially @devblackops and @RamblingCookieMonster, but with an effort to minimize the number of new dependencies for now so I haven't included the use of psake, or BuildHelpers/PowerShellBuild. In the future I think it would be nice to strategize on a tool / framework for better structuring discrete tasks during build/test.

While working on this, I found opportunities to improve the developer experience and docs as well, so the PR:

- Moves all source into an `ImportModule/` subfolder.
- Moves the readme and docs into a `docs/` subfolder.
- Adds a build step to "compile" a copy of the module in the `Output/` folder.
- Adds `RunTests.ps1` as an entrypoint to setup and run tests against the compiled module
- Adds platyPS to the build step to add/update markdown help files.
- Adds a devcontainer so that contributors can code in a local devcontainer or in a GitHub Codespace and have all the dependencies necessary to work on the module, including the docs which are served by mkdocs.
- Adds a docs GitHub Actions workflow to build and publish documentation to GitHub Pages.

I did not _intend_ to change behavior or add/remove functionality with the exception of reducing the module import time, but I recognize that the scope of this PR carries a higher risk of unintended side effects than a more typical PR. Thank you @dfinke for your time today to review the changes - hopefully we can get some customer feedback to confirm that there are no unintended side effects when used with real data!


dfinke:
@joshooaj Thanks for the work and PR. Going to pass on this. If the module continues to grow and folks are experiencing perf hits , it'll be a good time to leverage your work.

Thanks for the github action for the docs generation. Once that popped back on my radar I've seen several non-PowerShell projects using mkdocs. I will revisit it for ImportExcel and already working on implementing for another of my PowerShell modules.

Thanks again!

joshooaj:
No worries @dfinke - I appreciate you taking the time to chat about it and I'm glad some of it might be useful in other projects. Cheers!

PR #1562, opened 2024-02-05
Create ConditionalFormattingUnmatchColumns-Sol1.ps1

wyatt-wong:
Added new example for Conditional Formatting for Unmatch Columns


dfinke:
That’s great, thank you!

wyatt-wong:
> That’s great, thank you!

I was unable to find a solution to highlight the difference between 2 columns but I got the idea from the following example and I perform further some testing until I was able to come up with this solution.

[https://github.com/dfinke/ImportExcel/blob/ab526558a3cd643e10cad3c04f35e3645389d857/Examples/ConditionalFormatting/ConditionalFormattingIcontSetOnlyIcon.ps1](url)

wyatt-wong:
Noted

PR #1578, opened 2024-03-14
ChartDataSeparatePage.ps1

RipFence:
Added example script to  illustrate creating a chart on one excel worksheet when the data is on another worksheet.


dfinke:
Thanks @RipFence for the question and doing the PR!

PR #1601, opened 2024-05-17
[Import-Excel] Don't request ReadWrite access to the file

edwardmiller-mesirow:
Write access can cause locking issues, and I don't think it is really needed for this cmdlet.

Tested this locally and it works

### DIFF
--- a/Public/Import-Excel.ps1
+++ b/Public/Import-Excel.ps1
@@ -119,7 +119,7 @@
                     throw "'$($Path)' file not found"
                 }
 
-                $stream = New-Object -TypeName System.IO.FileStream -ArgumentList $Path, 'Open', 'Read', 'ReadWrite'
+                $stream = New-Object -TypeName System.IO.FileStream -ArgumentList $Path, 'Open', 'Read'
                 $ExcelPackage = New-Object -TypeName OfficeOpenXml.ExcelPackage
                 if ($Password) { $ExcelPackage.Load($stream, $Password) }
                 else { $ExcelPackage.Load($stream) }


dfinke:
Interesting, have not had a locking issue. Will investigate.

edwardmiller-mesirow:
I switched to this module because there was some terrible existing OLEDB-based powershell code we had that was suffering from locking issues. Which is especially unfortunate, given that the operations were read-only

Switching to this library actually solved that for the import step, but then a subsequent archive step was then running into locking issues from a file on a network drive.

It could literally be the case that the file was just open somewhere. But I also wanted to eliminate any possibility that a temporary file lock is not released.

I can see by searching this library for ReadWrite that the original code for this was added precisely to circumvent locking issues.

But, they seemed to think the ReadWrite permissions were what fixed that, and I think the usage of the stream is what fixed it and ReadWrite permissions are not necessary.

In terms of the principle of least privilege it just doesn't make sense to grant this process write permission to the file since it is not writing to it.

edwardmiller-mesirow:
In fairness, with the existing code I cannot locally replicate a locking issue either with or without this change. But, it does work with less privileges and that feels like a win.

edwardmiller-mesirow:
Here was the relevant related PR that I mentioned.

https://github.com/dfinke/ImportExcel/pull/35

nlsdg:
The ReadWrite in the code is for the FileShare enum. This parameter is to control how other processes can access the file.
It doesn't mean we request write access to the file. That is done through the FileAccess parameter (but it is set to Read).

If you go ahead with this change, you could run into problems.
1) Requests to open the file for writing by this or another process will fail until the FileStream object has been closed, but read attempts will succeed.
2) If another process already has the file open for writing the call will fail, because it tries to gain exclusive access. (which was the reason to change it in #35.)
Check the documentation on the System.IO.Filestream constructors.

PR #1602, opened 2024-05-18
[Import-Excel] Performance optimization

edwardmiller-mesirow:
This reduces redundant lookups on a per-cell basis.

### DIFF
--- a/Public/Import-Excel.ps1
+++ b/Public/Import-Excel.ps1
@@ -1,4 +1,4 @@
-﻿function Import-Excel {
+function Import-Excel {
     [CmdLetBinding()]
     [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingPlainTextForPassword", "")]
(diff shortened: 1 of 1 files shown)


dfinke:
Thanks, good optimiaztion, need to look at it. Need to look at the tests around it.

There is a mega perf issue I introduced when I added slick feature `'Import-Excel sales.xlsx *`.
It reads all the sheets and returns a hashtable keys are the sheet names and values are the data in the sheet.

Unfortunately how I implemented it if you are reading a large sheet, lots of columns and rows, it can go from 20 seconds to minutes. You can search the issues for when it was implemented. Folks pulled the prev version.

If you are so inclined

edwardmiller-mesirow:
Just pushed another commit to this branch that should help with that. I'm seeing 33% faster imports with a 1.4 megabyte spreadsheet. It goes to figure that a file with lots of columns would have a much greater percentage improvement.

PR #1603, opened 2024-05-18
[ConvertTo-ExcelXlsx] open XLS as read-only

edwardmiller-mesirow:
This could help prevent file locking issues. Also put everything in a try-finally to be extra sure that it will quit.

Just some extra guardrails.

see also: https://learn.microsoft.com/en-US/dotnet/api/microsoft.office.interop.excel.workbooks.open

I tested this locally and it works


dfinke:
Thanks, looks good. Like the extras on the Excel object model.

dfinke:
Published. Are you on the X platform? Let me know, when I post I can add. - Thanks again

PR #1604, opened 2024-05-20
update to .NET Standard 2.0 DLL for EPPlus

edwardmiller-mesirow:
The last version of the LGPL version of EPPlus was 4.5.3.3 and targets .NET Standard 2.0 (and .NET Standard 2.1)

The version currently on the master branch is 4.5.3.2 targeting .NET Framework 4

.NET Standard 2.0 supports .NET Framework 4.6 and above. Anything older is out of support anyways.

Ideally using .NET Standard 2.1 where possible would be even better.

Using .NET Standard 2.0+ could potentially allow newer APIs to be leveraged for EPPlus and thus improve performance.

I ran preliminary tests locally and it worked ok, as long as all the transitive dependencies were included as well.

https://www.nuget.org/packages/EPPlus/4.5.3.3


dfinke:
Why did you close?

edwardmiller-mesirow:
The builds were failing and I wasn't sure why. So I guess I'm giving up on this one for now.

I personally didn't see a very noticeable performance benefit with my test that I ran.

Was also considering using a step to pull these DLLs down from NuGet or something, instead of cluttering the repository with binary blobs. So I don't really like this PR very much right now even if it were working.

If you want to re-open and work on it, that's cool.

dfinke:
Thanks for looking into that. It is a single DLL that is why I put it in the repo. Didn't want to support another set of moving parts.

Yeah, the EPPlus folks did a great job on it and deserve to get paid for they decade+ work.

Wish we could use the new version.

PR #1605, opened 2024-05-21
[ConvertTo-ExcelXlsx] Accept a collection of paths

edwardmiller-mesirow:
If I loop through a big list of XLS files that I want to convert, it can throw an error at some point in the list:

```
Unable to get the SaveAs property of the Workbook class
```

This could be because of either file locking or issues performing round-trips to network locations. In a situation with network paths, it will make two full roundtrips per path entirely via Excel interop. And that could be a very error-prone situation. When converting the same list of files repeatedly, I was seeing it fail randomly each time on different files in that same list.

The changes in this PR:

1. This allows you to loop over a collection of paths that you want to convert, while only needing to create one COM object.

    - This is noticeably faster. Processing a dozen items for me goes from 2 minutes to 1 minute.

2. The $workbook object created by the Open method is used directly, instead of relying on the stateful ActiveWorkbook property of the $Excel object.

3. This optionally allows caching a copy of each file to the local Temp directory first, which can avoid both locking and network roundtrips. I made the CacheDirectory configurable as well, to handle the case where the default Temp directory isn't writeable. The default behavior is the same as before.

4. It also just uses ChangeExtension to set the extension since that will work a little better overall in cross-platform scenarios with different case sensitivity.

### DIFF
--- a/Public/ConvertTo-ExcelXlsx.ps1
+++ b/Public/ConvertTo-ExcelXlsx.ps1
@@ -47,12 +47,16 @@ function ConvertTo-ExcelXlsx {
 
         try {  
             $Excel.Visible = $false
-            $null = $Excel.Workbooks.Open($xlsFile.FullName, $null, $true)
-            $Excel.ActiveWorkbook.SaveAs($xlsxPath, $xlFixedFormat)
+            $workbook = $Excel.Workbooks.Open($xlsFile.FullName, $null, $true)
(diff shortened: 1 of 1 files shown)


edwardmiller-mesirow:
@dfinke this is ready and seems to be considerably better in the case of converting an entire collection of XLS files. But should still work exactly as before in the case of just a single path.

PR #1648, opened 2024-10-21
Remove flagged URL

evenmartinsen:
Remove old pesterbdd URL in Import-excel.psd1 due to conflicts with AV sollutions.


dfinke:
@evenmartinsen Thanks for using ImportExcel and fixing!

Issue #1459, opened 2023-06-01
Error when imported data starts with equals sign ("=")


An-dir:
I have noticed that there are problems importing data if the imported text starts with a "=". There may be a better solution than the current one.

```
#How I did my tests with unsatisfactory results
# Testresult: Errors
"Col1;Col2`r`ntext;=|0" | ConvertFrom-Csv -Delimiter ";" | Export-Excel test1.xlsx -AutoFilter -AutoSize
# Testresult: Displays the '-character
"Col1;Col2`r`ntext;'=|0" | ConvertFrom-Csv -Delimiter ";" | Export-Excel -AutoFilter -AutoSize
# works but with ugly interpretation as text in a formula
"Col1;Col2`r`ntext;=`"=|0`"" | ConvertFrom-Csv -Delimiter ";" | Export-Excel -AutoFilter -AutoSize
```

In an old issue you wrote ( https://github.com/dfinke/ImportExcel/issues/1007#issuecomment-798961673 ) that Excel interprets this as a formula, but the following test looks promising that there may be a better solution:
```
# Here is how I expect it to be handled (delimiter is language dependend in your test):
"Col1;Col2`r`ntext;=|0" | ConvertFrom-Csv -Delimiter ";" | Export-Csv test.csv -Delimiter ";" -NoTypeInformation
start excel test.csv
```
Then save as xlsx has no problem with displaying it as string and reopen the xlsx without errors. The recognition as a formula starts when you try to make changes - then you can decide what you want to do with the data.

There can be more characters than the "=" character or "'" at the beginning, wich can cause problems when importing. 

It looks like in the "sharedStrings.xml" is missing "<si><t>=|0</t></si>" and maybe some other errors, like counter in this and maybe other files.

Maybe you can fix it so that when you import, it will not crash, but handle it like the xlsx via csv did

dfinke:
please could you post a repro of the issue. data, script etc

saviourofdp:
A simple example that I have come across is using the output of _Compare-Object_ 

```ps

dfinke:
Yes. You need to handle that in your PowerShell script. Translate `==` to `equal` for example or prefix it with a`'` which is how native Excel enables you to start data with an `=` and it does not try to execute it as a function.

saviourofdp:
Unless I'm doing something wrong, the `'` solution is not the same. If I create a workbook manually with a field "Test" with the value "'=" and create a second workbook with the script
```ps
[...]

PR #1650, opened 2024-10-28
Add NoFormulaConversion to Export-Excel

AJ0hn:
This feature will allow formulas to be 'disabled' when using Export-Excel by treating the values as text.

This is especially useful when your data contains cells that may start with =, such as a Compare-Object output, without having to reformat the data. E.g. "equals". (#1459)

I've kept the naming convention and functionality similar to the existing 'overrides'. (NoNumberConversion and NoHyperLinkConversion)

### DIFF
--- a/Examples/NoFormulaConversion/NoFormula-AllColumns.ps1
+++ b/Examples/NoFormulaConversion/NoFormula-AllColumns.ps1
@@ -1,0 +1,13 @@
+try {Import-Module $PSScriptRoot\..\..\ImportExcel.psd1} catch {throw ; return}
+
+$xlSourcefile = "$env:TEMP\NoFormulaExample.xlsx"
+
+#Remove existing file
+Remove-Item $xlSourcefile -ErrorAction Ignore
+
+#These formulas should not calculate and remain as text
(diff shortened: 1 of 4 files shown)


dfinke:
@AJ0hn Thanks for the PR. I would need tests for this and I don't know off the top of my head if there are some in place for the other features of Export-Excel.

I'd also like an example script with data showing how the formulas are passed in. 

Thank you.

AJ0hn:
Thanks for the quick reply @dfinke .

Sure, I've added tests and examples as requested.

Please let me know if you have any questions. 👍

AJ0hn:
Hi @dfinke ,

Just wondering if you had any thoughts on this one?

We are currently using the below workaround to add an apostrophe where an equals sign appears at the beginning of a CSV cell.

```
Get-Content $csvPath | ForEach-Object {
    $_ -replace '(?<=^|,|,"|")=',"'="
}
```
Thanks!

dfinke:
Not going to have time to look this over, and don't know if this will be a breaking change.

The workaround looks straighforward and this is the first time it has been raised an an issue.

PR #1741, opened 2026-07-29
Fix Set-ExcelRange -Height changing the height of the row after the range

### LINKED ISSUE #1738: Set-ExcelRow -Height modifies the height of the next row as well
When using `Set-ExcelRow` or `Set-ExcelRange` to set the height of a row, the height of the next row is modified as well.

Steps to reproduce the issue:
```PowerShell
$Data = ConvertFrom-Csv @"
A,B,C
1,2,3
4,5,6
7,8,9
10,11,12
13,14,15
16,17,18
"@

$XlPkg = $Data | Export-Excel -WorksheetName TestHeight -TableStyle None -PassThru

Set-ExcelRow -Worksheet $XlPkg.TestHeight -Row 1 -Height 42 -WrapText -VerticalAlignment Top -HorizontalAlignment Center -Bold -Italic -Underline
Set-ExcelRange -Worksheet $XlPkg.TestHeight -Range '4:5' -Height 32 -WrapText -VerticalAlignment Top -HorizontalAlignment Center -Bold -Italic -Underline

foreach ($r in 1..$XlPkg.TestHeight.Dimension.Rows) { $XlPkg.TestHeight.Row($r).Height }
```
The output of the loop in the last line is:
```
42
42
15
32
32
32
15
```
The resulting Excel file looks like this:
Image

I would have expected that only the height of row 1 is set to 42, and only the height of rows 4 and 5 are set to 32, while the height of the other rows remains unchanged (i.e. the default 15).

What happens instead is that the height of rows 1 _and 2_ are set to 42, and the height of rows 4, 5 _and 6_ are set to 32.

Setting all those other formating properties is not necessary to trigger this issue.  I've only specified those options to demonstrate that they seem to behave as expected, i.e. they affect only the given rows.  Only the `-Height` "overflows" into the next row.
`Set-ExcelColumn -Width` seems to be working as expected as well.

Mike-Crowley:
Fixes #1738

## Problem

`Set-ExcelRange -Height` (and `Set-ExcelRow -Height`, which delegates to it) also changed the height of the first row *after* the requested range. From the issue's repro, setting row 1 to 42 and rows 4:5 to 32 produced:

```
42, 42, 15, 32, 32, 32, 15   # rows 2 and 6 wrongly modified
```

## Cause

An off-by-one in the row loop in `Set-ExcelRange`. EPPlus's `ExcelRange.Rows` is a row *count* (`End.Row - Start.Row + 1`), so

```powershell
($Range.Start.Row)..($Range.Start.Row + $Range.Rows)
```

iterates one row past the end of the range (a 1-row range loops `N..N+1`). The equivalent `-Width` logic for columns already subtracts 1, which is why `Set-ExcelColumn -Width` was unaffected.

## Fix

Subtract 1, matching the `-Width` branch:

```powershell
($Range.Start.Row)..($Range.Start.Row + $Range.Rows - 1)
```

The issue's repro now produces `42, 15, 15, 32, 32, 15, 15` as expected. As a side effect this also fixes a crash when the range ends at the worksheet's last possible row (the old code called `.Row(1048577)`, which throws).

## Tests

Added a regression context to `Set-Row_Set-Column-SetFormat.tests.ps1` covering `Set-ExcelRow` on a single row and `Set-ExcelRange` on a multi-row range, asserting the neighbouring rows keep the default height. The new tests fail against the unfixed code ("Expected 15, but got 42") and pass with the fix. Also added the same `Import-Module $PSScriptRoot\..\ImportExcel.psd1 -Force` line used by sibling test files so this file tests the repo code when run standalone.

Verified on Windows PowerShell 5.1 and PowerShell 7; full test file passes (22/22).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Public/Set-ExcelRange.ps1
+++ b/Public/Set-ExcelRange.ps1
@@ -159,7 +159,7 @@
             if ($PSBoundParameters.ContainsKey('Height')) {
                 if ($Range -is [OfficeOpenXml.ExcelRow]   ) {$Range.Height = $Height }
                 elseif ($Range -is [OfficeOpenXml.ExcelRange] ) {
-                    ($Range.Start.Row)..($Range.Start.Row + $Range.Rows) |
+                    ($Range.Start.Row)..($Range.Start.Row + $Range.Rows - 1) |
                         ForEach-Object {$Range.Worksheet.Row($_).Height = $Height }
                 }
                 else {Write-Warning -Message ("Can set the height of a row or a range but not a {0} object" -f ($Range.GetType().name)) }
--- a/__tests__/Set-Row_Set-Column-SetFormat.tests.ps1
+++ b/__tests__/Set-Row_Set-Column-SetFormat.tests.ps1
@@ -1,4 +1,4 @@
-﻿
+﻿Import-Module $PSScriptRoot\..\ImportExcel.psd1 -Force
 
 Describe "Number format expansion and setting" {
     BeforeAll {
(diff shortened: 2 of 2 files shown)


Mike-Crowley:
Checking in, no rush at all: just making sure nothing here is waiting on me. Happy to rebase, tweak, or split anything up if that makes review easier.

One small thing I noticed: Copilot opened #1746 for this same off-by-one. Its one-line change is identical to this PR, and #1741 also carries the regression tests (single-row and multi-row) plus a fix for the crash when a range ends on the worksheet's last row, so this one is a superset. Whichever you prefer to take is totally fine by me; just flagging so the two don't get merged on top of each other.

